PR #10, opened 2017-12-13
Don't run theia as root

JanKoehnlein:


hexa00:
@JanKoehnlein  I think we need a startup script and run a chown theia:theia /home/project for this to work :(

JamesDrummond:
@hexa00 Maybe provide a separate docker image/Dockerfile that runs as normal user. Take a look at this post for a conversation on the issue https://github.com/moby/moby/issues/2259 in regards to the volume binding. You can check/test the '/home/project' folder for correct permissions in Theia application then prompt user to change permissions if not set correct.

You can also do what Che did a provide a base image that different images can use https://github.com/eclipse/che/blob/master/dockerfiles/base/Dockerfile . So for no root Dockerfile  it would just be:

```
FROM theiaide/theia
RUN useradd -m theia
USER theia
ENTRYPOINT [ "npx", "theia", "start", "/home/project", "--hostname=0.0.0.0" ]
```

JamesDrummond:
I have a PR at https://github.com/theia-ide/theia-apps/pull/41 that could cover this.

PR #18, opened 2018-01-05
[theia-docker] use alpine node image as a base + get rid of electron

akosyakov:


akosyakov:
@JanKoehnlein they need to be tested for missing libraries, alpine image is very slim

We can do it separately.

PR #19, opened 2018-01-05
Get rid of python, make and other build time libraries

akosyakov:


akosyakov:
I'm merging it

hexa00:
You removed bash but ENV SHELL is still /bin/bash ?

hexa00:
Oops I missread the diff... disregard last comment

PR #31, opened 2018-03-22
added typescript

akosyakov:


akosyakov:
@JanKoehnlein ok, will do

PR #35, opened 2018-04-01
Theia IDE on user-mode

### LINKED ISSUE #34: Make Theia images run on usermode
theia images should run on user mode via ``USER`` + a ``RUN`` command that adds the user.

There are a couple of factors why it should be done like this:
- OpenShift compatibility/k8s compatibility
- this is how other Cloud IDEs like Cloud9 and Codenvy/Eclipse Che author their workspace stacks.

sr229:
Resolves #34  and closes #10 as this supersedes the following.

Runs images in `theia` user and fixes file permissions to be compatible with OpenShift/K8s.

### DIFF
--- a/theia-docker/Dockerfile
+++ b/theia-docker/Dockerfile
@@ -13,13 +13,13 @@ FROM node:8-alpine
 RUN addgroup theia && \ 
     adduser -G theia -s /bin/sh -D theia;
 RUN chmod g+rw /home && \
-    mkdir /home/theia/project && \
+    mkdir -p /home/project && \
     chown theia:theia /home/theia/project;    
 RUN apk add --no-cache git openssh bash
-WORKDIR /home/theia/project
+WORKDIR /home/project
 COPY --from=0 /home/theia /home/theia
 EXPOSE 3000
 ENV SHELL /bin/bash
 ENV USE_LOCAL_GIT true
 USER theia
-ENTRYPOINT [ "yarn", "theia", "start", "/home/theia/project", "--hostname=0.0.0.0" ]
+ENTRYPOINT [ "yarn", "theia", "start", "/home/project", "--hostname=0.0.0.0" ]
--- a/theia-docker/Dockerfile
+++ b/theia-docker/Dockerfile
@@ -14,7 +14,7 @@ RUN addgroup theia && \
     adduser -G theia -s /bin/sh -D theia;
 RUN chmod g+rw /home && \
     mkdir -p /home/project && \
-    chown theia:theia /home/theia/project;    
+    chown theia:theia /home/project;    
 RUN apk add --no-cache git openssh bash
 WORKDIR /home/project
 COPY --from=0 /home/theia /home/theia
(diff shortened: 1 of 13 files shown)


epatpol:
@akosyakov WDYT?

sr229:
@epatpol looks like its not possible on my end.

JanKoehnlein:
Did you have a look at the discussion in https://github.com/theia-ide/theia-apps/pull/10, i.e. did you consider the file permissions of a mounted volume?

sr229:
@JanKoehnlein this is verified using OpenShift volumes. If volume has no proper permissions, it would cause a back-off, so yes

svenefftinge:
@sr229 You need to verify it with plain docker mount as described in the readme.

sr229:
@svenefftinge verified to mount properly

theia-docker/Dockerfile
```diff
@@ -9,10 +9,17 @@ RUN yarn && \
...
 RUN apk add --no-cache git openssh bash
-WORKDIR /home/theia
+WORKDIR /home/theia/project
 COPY --from=0 /home/theia /home/theia
 EXPOSE 3000
 ENV SHELL /bin/bash
 ENV USE_LOCAL_GIT true
-ENTRYPOINT [ "yarn", "theia", "start", "/home/project", "--hostname=0.0.0.0" ]
+USER theia
+ENTRYPOINT [ "yarn", "theia", "start", "/home/theia/project", "--hostname=0.0.0.0" ]
```
akosyakov: Is there a way to preserve `/home/project` that we don't need to updated repo, images descriptions, tweets and so on?

akosyakov: `mkdir home/project`? or i am missing something

sr229: Will add that instead of doing it on the default user folder

colemickens:
What's the status on this? It's a good change to see...

sr229:
This PR's build seems to be affected by a 429 from NPM, called a rebuild by sending a empty commit

marcdumais-work:
@sr229 This is a known issue that happens when too many Travis builds, probably from a given ip address, request resources from GH. In our case it's the ripgrep npm module that, upon installation, downloads a native executable for the current platform. We have done what we could to mitigate this, but it can still happen. 

I have restarted your build - hopefully it's less busy now and will pass.

sr229:
There's still some errors but hopefully this aren't Ratelimited messages and actual errors

sr229:
@marcdumais-work I built them all in [Play With Docker](https://labs.play-with-docker.com) and they seem to work now

akosyakov:
https://github.com/theia-ide/theia-apps/pull/35/files#diff-3b118c36064586431ebe83d500e97cf3 looks good to me, @marcdumais-work if everything else looks good then @sr229 can squash commits and merge?

sr229:
Yeah it seems I can't squash it on my end, needs to be done on a merge

sr229:
Closing this to do another PR to clean this up

PR #37, opened 2018-04-12
Use --password-stdin instead of --password

akosyakov:
It should fix the build: https://travis-ci.org/theia-ide/theia-apps/builds/365328452

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ install:
   - docker tag "$IMAGE_TAG" "$IMAGE:$NPM_TAG"
 
 before_deploy:
-  - docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -35,7 +35,7 @@ install:
   - docker tag "$IMAGE_TAG" "$IMAGE:$NPM_TAG"
 
 before_deploy:
-  - docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
+  - docker login -u "$DOCKER_USERNAME" --password-stdin "$DOCKER_PASSWORD"
```
JanKoehnlein: shouldn't it be

    echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin

if it should be taken from stdin?

akosyakov: thanks, changed

PR #50, opened 2018-07-05
theia-full-docker: include Rust + rls

### LINKED ISSUE #48: Theia Full does not contain the Rust plugin

colemickens:
Fixes #48.

Blocked on https://github.com/theia-ide/theia-rust-extension/issues/2

### DIFF
--- a/theia-full-docker/Dockerfile
+++ b/theia-full-docker/Dockerfile
@@ -110,6 +110,13 @@ RUN ln -s /usr/bin/clangd-6.0 /usr/bin/clangd
 RUN apt-get update && apt-get install -y python python-pip
 RUN pip install python-language-server
 
+#Rust
+RUN curl https://sh.rustup.rs -sSf > /tmp/rustup.sh
+RUN chmod +x /tmp/rustup.sh && /tmp/rustup.sh -y
+ENV PATH=$PATH:/root/.cargo/bin
+RUN rustup update
+RUN rustup component add rls-preview rust-analysis rust-src
+
 #Theia
 ##Needed for node-gyp, nsfw build
 RUN apt-get update && apt-get install -y python build-essential
--- a/theia-full-docker/Dockerfile
+++ b/theia-full-docker/Dockerfile
@@ -1,7 +1,7 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 #Common deps
-RUN apt-get update && apt-get -y install curl xz-utils
+RUN apt-get update && apt-get -y install curl xz-utils gnupg
 
 #Install node and yarn
 #From: https://github.com/nodejs/docker-node/blob/6b8d86d6ad59e0d1e7a94cec2e909cad137a028f/8/Dockerfile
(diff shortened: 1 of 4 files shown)


theia-full-docker/Dockerfile
```diff
@@ -110,6 +110,13 @@ RUN ln -s /usr/bin/clangd-6.0 /usr/bin/clangd
 RUN apt-get update && apt-get install -y python python-pip
 RUN pip install python-language-server
 
+#Rust
+RUN curl https://sh.rustup.rs -sSf > /tmp/rustup.sh
+RUN chmod +x /tmp/rustup.sh && /tmp/rustup.sh -y
```
marcdumais-work: When building the image on my Ubuntu 16.04 machine, I get the following error:
```
Step 41/53 : RUN chmod +x /tmp/rustup.sh && /tmp/rustup.sh -y
 ---> Running in 73694bf1a4f1
/bin/sh: 1: /tmp/rustup.sh: Text file busy
The command '/bin/sh -c chmod +x /tmp/rustup.sh && /tmp/rustup.sh -y' returned a non-zero code: 2

```

marcdumais-work:
Please sign your commit:
```
git commit --amend -s
```

theia-full-docker/Dockerfile
marcdumais-work: Replacing the line with ```RUN /bin/sh /tmp/rustup.sh -y``` seems to work for me.

theia-full-docker/latest.package.json
```diff
@@ -21,6 +21,7 @@
         "@theia/preferences": "latest",
         "@theia/process": "latest",
         "@theia/python": "latest",
+        "@theia/rust": "latest",
```
marcdumais-work: It looks like @theia/rust is not yet published on npm, correct?

theia-full-docker/Dockerfile
colemickens: I don't understand how there could be a difference here, unless something changed in the base image and you have an older base image cached.

I'm using this Dockerfile currently, without issue. Further, I can't imagine a script that `/bin/sh` could run that `/bin/bash` (presumably the default in Ubuntu) couldn't run.

marcdumais-work: I think the issue I encountered may be this one: https://github.com/moby/moby/issues/9547

Nothing wrong as such with your version - technically it should work just as well. However the issue I reported happened consistently on my work laptop. The same Dockerfile builds fine on my (much faster) home desktop. According to the link above, it may be linked to the ```aufs``` storage driver. I do not remember if that's used on my work laptop, but it's not on my desktop, so that may be a factor.

>unless something changed in the base image and you have an older base image cached

This image is self-contained and doesn't derive from the base image. I used ```--no-cache``` when building to make sure it was a clean build.

>I can't imagine a script that /bin/sh could run that /bin/bash (presumably the default in Ubuntu) couldn't run.

looks like docker is using /bin/sh, so I used the same. from the error message: ```/bin/sh: 1: /tmp/rustup.sh: Text file busy```. My goal was to avoid having to use chmod before executing the file, thus reducing the two commands to one. 



Using a ```sync``` between the ```chmod``` and script execution may be an alternative: https://github.com/moby/moby/issues/9547#issuecomment-77547893

This user has a theory about what's happening: https://github.com/moby/moby/issues/9547#issuecomment-337684598

PR #58, opened 2018-07-09
fix #38: build one dockerfile per a stage

### LINKED ISSUE #38: GH API rate error
We have too many jobs running in parallel, one GH token is not enough. We should rather use several as secrets in travis file or break down jobs to several stages to avoid running all them together.

akosyakov:


akosyakov:
I don't think anything can be done to fix failures for PRs. The issue is that Travis does not provide access to secure env variables for PRs. They will fail always.

Should we try this PR for master? The issue with stages that failing of one stage prevents running next stage. But since it should reduce the number of simultaneous requests maybe there won't be failing stages.

PR #73, opened 2018-07-28
add theia/json

jgroom33:
#63

### DIFF
--- a/theia-python-docker/README.md
+++ b/theia-python-docker/README.md
@@ -3,30 +3,17 @@
 ### Build latest
 
 ```bash
-docker build . -t theiaide/theia-python:latest
+docker build . -t theia-python:latest
 ```
 
 ### Build next
 
 ```bash
-docker build --build-arg version=next . -t theiaide/theia-python:next
+docker build --build-arg version=next . -t theia-python:next
 ```
 
 ### Run locally on Linux or OS X
 
 ```bash
-docker run -it -p 3000:3000 -v "$(pwd):/home/project" theiaide/theia-python:latest
-```
-
-### Push to Docker Hub
-
-```bash
-docker login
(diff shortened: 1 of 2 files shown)


theia-python-docker/README.md
```diff
@@ -3,31 +3,17 @@
 ### Build latest
 
 ```bash
-docker build . -t theiaide/theia-python:0.3.12
-docker build . -t theiaide/theia-python:latest
+docker build . -t theia-python:latest
```
marcdumais-work: Should it me mentioned that it might be necessary to change the value of variable "version" in DockerFile to match the wanted version: latest or next?

jgroom33: @marcdumais-work This is noted as well: `docker build --build-arg version=next . -t theiaide/theia-python:next`

marcdumais-work: @jgroom33 I meant that specifying a version in the docker build command is not enough to actually have it use that version of Theia, when building the image. One also needs to specify ```latest``` or ```next``` in Dockerfile, variable ```version```.  But that's the case for other images too, so maybe no need to mention.

marcdumais-work:
@jgroom33 to address #63, the PR would need to add the ```@theia/json``` extension to the other images too. Is that something you'd like to include in this PR?

marcdumais-work:
I am not able to build the ```latest``` version of the python image from this PR (webpack error). I think it's because we do not yet have a ```@theia/json : latest``` published on npm, yet. 

With tomorrow's Theia release, that should no longer be a problem. Will re-try then.

vince-fugnitto:
Closed thanks to https://github.com/theia-ide/theia-apps/commit/7bc43b199f1b144dcc3510ed4eec29543d133cec
`@theia/json` has been added to all docker images.

Issue #77, opened 2018-09-03
Create a Docker image that can build electron-based Theia applications for Windows on Linux


kittaakos:
The image should be published under [theiaide](https://hub.docker.com/u/theiaide/) with `theia-electron-builder` name.

PR #78, opened 2018-09-03
GH-77: Added image to building electron applications for Windows on UNIX

kittaakos:
Closes: #77.

Signed-off-by: Akos Kitta <kittaakos@typefox.io>

### DIFF
--- a/theia-electron-builder/Dockerfile
+++ b/theia-electron-builder/Dockerfile
@@ -1,0 +1,51 @@
+FROM electronuserland/builder:wine
+
+# Replace shell with bash so we can source files.
+RUN rm /bin/sh && ln -s /bin/bash /bin/sh
+
+RUN apt-get update -q --fix-missing \
+# Install `add-apt-repository`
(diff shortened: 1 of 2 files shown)


akosyakov:
@kittaakos do you want to add it to travis?

kittaakos:
Good question; I would say no. There is not much to build, and it does not depend on Theia at all. In a nutshell: it is not subject to change. What do you think, @akosyakov?

kittaakos:
> don't forget to propagate a GH token properly

Can you please explain why do we need it?

akosyakov:
@kittaakos if it is not a part of travis then it does not matter. We need to avoid hitting the GH request rate limit by Travis while downloading vscode-rigrep.

kittaakos:
> GH request rate limit by Travis while downloading vscode-rigrep

That explains. Thank you! I am merging this PR. The Docker image is available [here](https://hub.docker.com/r/theiaide/theia-electron-builder/).

Issue #3, opened 2017-11-14
Add Theia desktop apps


akosyakov:
Latest and next theia-desktop app should be added. The auto build should be provided whenever a new @theia/core latest or next version is published.

kittaakos:
Just for the sake of clarity; did you mean an electron-based application?

akosyakov:
Yes, similar to what you did for yangster.

PR #79, opened 2018-09-03
GH-3: Added the electron-based Theia application.

kittaakos:
Closes: #3.

Bare minimum configuration to build the electron-based Theia application. So that you can use it for the development of Theia...

Signed-off-by: Akos Kitta <kittaakos@typefox.io>


kittaakos:
Related: https://github.com/theia-ide/theia/issues/3297

kittaakos:
It is ready for review. I need to check it once more on Windows too, but it works on OS X as an installed native application.

kittaakos:
> I need to check it once more on Windows too

🎉 It works.

kittaakos:
I do not know what is causing the build failure, but most likely it is unrelated to my changes.

[`master`](https://travis-ci.org/theia-ide/theia-apps/builds/457215860):

screen shot 2018-11-20 at 12 56 51

[`GH-3`](https://travis-ci.org/theia-ide/theia-apps/builds/457359438):

screen shot 2018-11-20 at 13 12 18

kittaakos:
@akosyakov, can you please look into this PR?

theia-electron/package.json
```diff
@@ -0,0 +1,64 @@
...
+        "@theia/navigator": "next",
+        "@theia/outline-view": "next",
+        "@theia/output": "next",
+        "@theia/preferences": "next",
+        "@theia/preview": "next",
+        "@theia/process": "next",
+        "@theia/search-in-workspace": "next",
+        "@theia/task": "next",
+        "@theia/terminal": "next",
+        "@theia/typescript": "next",
```
akosyakov: @kittaakos we expect that typescript is already installed on the machine? Can we bundle one version as a fallback?

kittaakos: I am not against bundling it, but it is not necessary. https://github.com/theia-ide/theia/issues/3568

marcdumais-work:
> I do not know what is causing the build failure, but most likely it is unrelated to my changes.

+1. It looks to be a yangster issue; probably it needs updating for some Theia API change that has happened since the last release.

theia-electron/package.json
```diff
@@ -0,0 +1,64 @@
+{
+    "private": true,
+    "name": "theia-electron",
+    "version": "0.3.16",
```
marcdumais-work: When I build for Linux, I get a .deb package: `theia-electron_0.3.16_amd64.deb`. I assume that the `0.3.16` part of the Debian package name comes from here? The same version is used in the metadata of the package:

image

It would be desirable for the exact `next` version to be used instead, since that version is used to determine if a given .deb package is newer than the one currently installed.

marcdumais-work:
Very nice. I tried building/installing the DEBIAN package generated for my Ubuntu machine, and it works nicely (or at least about as nice as if I build it from source).

theia-electron/package.json
kittaakos: Let's do it. I do not mind.

kittaakos:
I have added `typescript` `latest` as a dependency, did the last build locally, and verified the bundled application on macOS. I could do TS development in Theia. The build failures for this PR are identical with the build failures from the `master` (CRON). I am going to merge this...

`GH-3`:
screen shot 2018-12-03 at 10 42 31

`master`:
screen shot 2018-12-03 at 10 42 21

Issue #20, opened 2018-01-05
Reduce size of java image


akosyakov:
See for the guidance: https://github.com/theia-ide/theia-apps/issues/14#issuecomment-355573827

sr229:
Is it possible we can try OpenJDK images since we're using a more obsolete image for Java?

vince-fugnitto:
AFAIK, the size of the Java image is still an issue.

sr229:
@vince-fugnitto I don't think we can find a way to lower that since JDK itself is a big chonk of a SDK.

vince-fugnitto:
> @vince-fugnitto I don't think we can find a way to lower that since JDK itself is a big chonk of a SDK.

@sr229 do you propose the issue to be closed then?

vince-fugnitto:
> @vince-fugnitto I don't think we can find a way to lower that since JDK itself is a big chonk of a SDK.

Closing based on comment.

PR #90, opened 2018-10-09
Convert Java Image to OpenJDK8 and attempt to lower image size

sr229:
Fixes theia-ide/theia-apps#20


sr229:
@marcdumais-work seems to be an issue on `COPY`. Let me try to fix it

sr229:
@marcdumais-work Seems like the error this time is CI-wise, otherwise the Docker image now builds properly

sr229:
Probably ready to merge a this point @marcdumais-work. It works fine as we speak.

marcdumais-work:
Ok, thanks @sr229. I will give it a try building/testing the PR locally.

theia-java-docker/Dockerfile
```diff
@@ -26,7 +26,8 @@ ARG version=latest
 ADD $version.package.json ./package.json
 ARG GITHUB_TOKEN
 RUN yarn --cache-folder ./ycache && rm -rf ./ycache
-RUN yarn theia build
+RUN yarn theia build && yarn cache clean
+RUN apt-get clean
```
marcdumais-work: I do not think this does much good. IIRC for this to be worthwhile (i.e. save space), it would need to happen in the same ```RUN``` command as the one that created/updated the apt cache, else I think these files will still be present on that image layer, and just marked as removed in one of the following layers. 

So the "apt clean" command should probably be added to the "RUN apt-get update && ..." command on line 10.

sr229: Should be fixed on the latest change.

marcdumais-work:
@sr229 When I compare the size of the image I obtain when I build from this PR vs master, it seems that your version is bigger:
```
REPOSITORY            TAG                 IMAGE ID            CREATED             SIZE

theia-java-sr229      next                ec9ebab796c5        2 minutes ago       1.7GB
theia-java            next                8adbbaea3a88        9 minutes ago       1.39GB
```
Can you confirm if you see the same or similar? This at least partly defeats the stated purpose of this PR (the commit only mentions the "slimming" aspect). 

About going with JDK 11, it may still be a little early.

Our Java LS is "jdt.ls", which is  based on Eclipse jdt. It looks like JDK 11 support has only recently been added:  https://github.com/eclipse/eclipse.jdt.ls/issues/735 . It's not clear if a build containing this is available yet. And I think ATM we have our own jdt.ls build for Theia, instead of using what's released officially by the project. So more work may be needed. 

note: looks like JDK 10 also not [working](https://github.com/eclipse/eclipse.jdt.ls/issues/740)

sr229:
@marcdumais-work would JDK8/9 suffice for now?

marcdumais-work:
> would JDK8/9 suffice for now?

Yes, I think so. As soon as we can confirm that JDK 11 works well with our LS, then we should update.

sr229:
@marcdumais-work latest squash change should be JDK8 now

vince-fugnitto:
I think it's worthwhile to give the PR another test :)
@sr229 if you're still interested, do you mind rebasing the PR and fixing the conflicts so we can continue with the review?

sr229:
Lemme see if I can rebase it today via Gitpod @vince-fugnitto

vince-fugnitto:
Closing based on the comment https://github.com/theia-ide/theia-apps/issues/20#issuecomment-513761401

PR #99, opened 2018-11-01
[theia-docker] .yarnclean file

uniibu:
This is in relation with https://github.com/theia-ide/theia-apps/pull/94 and in reference to https://github.com/eclipse/che/pull/11765

Separating the .yarnclean file makes it easier to add new files/folders to it. 

Also removing unused modules since they are already bundled to bundle.js and bundle.js.map files.

Image size is around 210MB

Signed-off-by: Uni Sayo <unibtc@gmail.com>

### DIFF
--- a/theia-docker/.yarnclean
+++ b/theia-docker/.yarnclean
@@ -1,0 +1,77 @@
+# test directories
+__tests__
+test
+tests
+powered-test
+
+# asset directories
+docs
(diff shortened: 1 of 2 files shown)


uniibu:
Closing as this might break future builds. Having similar function integrated internally to theia is a much safer choice..

PR #101, opened 2018-11-08
Python3 image

### LINKED ISSUE #74: Support of Python3
It's great to see the Docker for Python added. I've tried it and it works well. However, it is for Python2 only currently. Can we add support of Python3?

sr229:
Resolves #74 

Adds a Python 3-specific image.

### DIFF
--- a/theia-python3-docker/Dockerfile
+++ b/theia-python3-docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8-stretch
+FROM node:10-stretch
 
 RUN apt-get update \
     && apt-get install -y python3 python3-dev python3-pip \ 
--- a/theia-python3-docker/Dockerfile
+++ b/theia-python3-docker/Dockerfile
@@ -1,0 +1,21 @@
+FROM node:8-stretch
+
+RUN apt-get update \
+    && apt-get install -y python3 python3-dev python3-pip \ 
+    && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/*
+
+RUN pip3 install \
+    python-language-server \
+    flake8 \
+    autopep8
+
+ARG version=latest
+
+WORKDIR /home/theia
+ADD $version.package.json ./package.json
+ARG GITHUB_TOKEN
+RUN yarn --cache-folder ./ycache && rm -rf ./ycache
+RUN yarn theia build
+EXPOSE 3000
+ENV SHELL /bin/bash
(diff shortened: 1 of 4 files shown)


sr229:
Could've used and copy out the Python3 image in DockerHub but it seems they compile it from source, may cause unecessary image sizing problems.

jgbradley1:
Should update the base image to Node 10 now.

jgbradley1:
Auto-completion and suggestions are not working for me when I try this image. I get the following error message when running the docker image and create a test python file.

```
root ERROR Python: /usr/bin/python: No module named pyls
```

It install expects python-language-server to be installed for Python 2.7. One workaround I tried was to put `alias python=python3` in the Dockerfile but it does not work as expected. Also using alias is the wrong way to do it according to the discussions [here](https://askubuntu.com/questions/320996/how-to-make-python-program-command-execute-python-3).

marcdumais-work:
Hi @sr229 , @jgbradley1,

Thanks for your continued work and validation of this PR. 

I wanted to bring to your attention a new opportunity: Theia is increasingly able to re-use VS Code extensions as-is. I do not know if anyone has tried the [Python](https://marketplace.visualstudio.com/items?itemName=ms-python.python) one yet, but it's likely to work well, and if not we can work in the main Theia repo to add any missing VS Code APIs. 

As an example, here's the corresponding PR for [golang](https://github.com/theia-ide/theia-apps/pull/146), where we replaced the Theia extension by the VS Code one.

jgbradley1:
@marcdumais-work @sr229 

I worked up a quick example with the [Python](https://marketplace.visualstudio.com/items?itemName=ms-python.python) extension and removed `@theia/python`. It works quite well out of the box. I see no immediate error messages on startup. A few issues exist with debugging and importing/running jupyter notebooks.

marcdumais-work:
> @marcdumais-work @sr229
> 
> I worked up an [example](https://github.com/jgbradley1/theia-python3) with the [Python](https://marketplace.visualstudio.com/items?itemName=ms-python.python) extension and removed `@theia/python`. It works quite well out of the box. I see no immediate error messages on startup. A few issues exist with debugging and importing/running jupyter notebooks.

Nice! About debugging: I am not sure about Python, but often debuggers use `ptrace` and so require the container to be run with special permssions. Not sure if you may have tried something like e.g. [here](https://github.com/theia-ide/theia-apps/blob/master/theia-go-docker/README.md): 
 `docker run --security-opt seccomp=unconfined ...`?

PR #103, opened 2018-11-21
[theia-docker] trim down docker image size

uniibu:
This is in relation to the merged pr https://github.com/theia-ide/theia-apps/pull/94 which seems to be mysteriously missing from the commit history.


benoitf:
I've pushed original commit so we could close this PR

PR #121, opened 2019-01-21
[full] update clangd and java JDK

marcdumais-work:
Switched to openjdk from Oracle jdk
Use latest development clangd-9


marcdumais-work:
Confirmed the docker-full build got ok pass the part with JDK and clangd. Merging.

PR #133, opened 2019-02-17
decrease number of layers (run)

avelino:


avelino:
upgrade in https://github.com/theia-ide/theia-apps/pull/136

PR #141, opened 2019-03-06
 go: upgrade go tools packages

avelino:

PR #117, opened 2019-01-15
upgrade go to go1.11

lanrat:
fixes #116

PR #146, opened 2019-03-11
[theia-go] migrate next to vscode extension

### LINKED ISSUE #118: golang version - any plans including delve debugger?
Do you have any plans including delve debugger for theia-go docker version?
https://github.com/go-delve/delve/releases

I am asking without knowing if it even possible to use a debugger with Theia?

akosyakov:
In order to start one should use:
```
docker run --security-opt seccomp=unconfined -it -p 3000:3000 -v "$(pwd):/home/project:cached" theiaide/theia-go:next
```

It allows ptrace for debugging. When you try it is important to install the project within docker container if you host OS is different.

Debugging https://github.com/demo-apps/go-gin-app:
Screen Shot 2019-03-12 at 14 16 33

Closes #141, fix #118

### DIFF
--- a/theia-go-docker/next.package.json
+++ b/theia-go-docker/next.package.json
@@ -1,22 +1,22 @@
 {
     "private": true,
     "dependencies": {
+        "@theia/plugin-ext-vscode": "next",
+        "@theia/preview": "next",
         "@theia/editor-preview": "next",
-        "@theia/go": "next",
-        "@theia/messages": "next",
-        "@theia/navigator": "next",
-        "@theia/terminal": "next",
-        "@theia/outline-view": "next",
-        "@theia/preferences": "next",
         "@theia/getting-started": "next",
-        "@theia/git": "next",
-        "@theia/file-search": "next",
-        "@theia/markers": "next",
-        "@theia/search-in-workspace": "next",
         "@theia/json": "next",
         "@theia/textmate-grammars": "next"
     },
     "devDependencies": {
-        "@theia/cli": "next"
+        "@theia/cli": "next",
+        "@theia/debug": "next"
+    },
+    "scripts": {
+        "postinstall": "download-debug-adapters"
+    },
+    "adapterDir": "plugins",
+    "adapters": {
+        "vscode-go": "https://ms-vscode.gallery.vsassets.io/_apis/public/gallery/publisher/ms-vscode/extension/Go/0.9.2/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage"
     }
 }
--- a/theia-go-docker/Dockerfile
+++ b/theia-go-docker/Dockerfile
@@ -1,41 +1,78 @@
 FROM node:8
 ARG version=latest
-ENV GO_VERSION 1.12
-RUN apt-get update && \
-    apt-get install -y curl && \
-    curl -sS https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar.gz | tar -C /usr/local -xzf -
-ENV PATH $PATH:/usr/local/go/bin
 WORKDIR /home/theia
 ADD $version.package.json ./package.json
 ARG GITHUB_TOKEN
-RUN yarn --cache-folder ./ycache && rm -rf ./ycache && \
-    yarn theia build
+RUN yarn --pure-lockfile && \
+    NODE_OPTIONS="--max_old_space_size=4096" yarn theia build && \
+    yarn --production && \
+    yarn autoclean --init && \
+    echo *.ts >> .yarnclean && \
+    echo *.ts.map >> .yarnclean && \
+    echo *.spec.* >> .yarnclean && \
+    yarn autoclean --force && \
+    rm -rf ./node_modules/electron && \
+    yarn cache clean
+
 # See : https://github.com/theia-ide/theia-apps/issues/34
 RUN adduser --disabled-password --gecos '' theia && \
     chmod g+rw /home && \
     mkdir -p /home/project && \
+    mkdir -p /home/go && \
+    mkdir -p /home/go-tools && \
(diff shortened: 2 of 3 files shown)


theia-go-docker/next.package.json
```diff
@@ -1,22 +1,22 @@
...
-        "@theia/cli": "next"
+        "@theia/cli": "next",
+        "@theia/debug": "next"
+    },
+    "scripts": {
+        "postinstall": "download-debug-adapters"
+    },
+    "adapterDir": "plugins",
+    "adapters": {
+        "vscode-builtin-go": "https://registry.npmjs.org/@theia/vscode-builtin-go/-/vscode-builtin-go-0.1.0.tgz",
```
akosyakov: should be removed after https://github.com/theia-ide/theia/pull/4547 gets merged

akosyakov:
Are they any objections to merge it? Could someone try?

theia-go-docker/next.package.json
```diff
@@ -1,22 +1,21 @@
 {
     "private": true,
     "dependencies": {
+        "@theia/plugin-ext-vscode": "next",
```
marcdumais-work: for your consideration: @theia/preview would be useful to have, to render Markdown. I guess that would be true for all images

marcdumais-work:
@akosyakov  I gave this PR a quick test, and the VS Code extension is pretty awesome. It works well, as far as I can tell. I noticed a few nice features we did not have before. I miss a bit the codelens for method references, but I'll live with the new one fine. Very nice that this works so well in Theia. I have not tried debug yet. 

```
> docker run --security-opt seccomp=unconfined -it -p 3000:3000 -v "$(pwd):/home/project:cached" theiaide/theia-go:next
```

I am not a go developer, but I wonder if this image is best still used like this, mounting a directory on the host in the container on directory`/home/project`, which is also used as `GOPATH`. 

Contributor  @lanrat recently had an interesting comment on this: 
https://github.com/theia-ide/theia-apps/pull/117#issuecomment-456206328

This could be considered separately from this PR.

akosyakov:
@lanrat @marcdumais-work It sounds like putting projects in GOPATH does not hurt projects using go modules and at the same time provides backward compatibility for projects not using modules. So mounting in GOPATH by default cover all use cases. Do i miss something?

marcdumais-work:
@akosyakov Testing debug: I have started the image with the suggested parameters and cloned the suggested go repo (go-gin-app) and built it (`go build`). I created a launch config but can't get the debugger to start: 

``` json
{
  "version": "0.2.0",
  "configurations": [

    {
      "name": "Launch",
      "type": "go",
      "request": "launch",
      "mode": "debug",
      "program": "${workspaceFolder}/go-gin-app",
      "env": {},
      "args": []
    }
  ]
}
```

When I try to launch the debugger using that config, I get an error pop-up: `Failed to continue: "{e}"` . Can you share your debug config?

marcdumais-work:
Exception when trying to start debugger (might be due to bad config):
```
root ERROR [hosted-plugin: 2617] (node:2617) UnhandledPromiseRejectionWarning: Error: the debug session connection is disposed, id: 7bf0bcf9-608c-42e3-9779-dc7affecd2bd
    at e.checkDisposed (http://localhost:3009/bundle.js:30:588550)
    at e.newEmitter (http://localhost:3009/bundle.js:30:591920)
    at e.getEmitter (http://localhost:3009/bundle.js:30:591808)
    at e.doFire (http://localhost:3009/bundle.js:30:591717)
    at e.handleEvent (http://localhost:3009/bundle.js:30:591486)
    at e.handleMessage (http://localhost:3009/bundle.js:30:590709)
    at t.callback (http://localhost:3009/bundle.js:30:589097)
    at t.readMessage (http://localhost:3009/bundle.js:30:753980)
    at e.<anonymous> (http://localhost:3009/bundle.js:79:1101881)
    at http://localhost:3009/bundle.js:79:1101396

root ERROR [hosted-plugin: 2617] (node:2617) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
```

akosyakov:
>  cloned the suggested go repo (go-gin-app)

In Go, one does not clone, but runs `go get github.com/demo-apps/go-gin-app` to install proper dependencies. It's better to run in docker container if your host OS is different, otherwise you can get different dependencies and tooling won't work.

theia-go-docker/next.package.json
akosyakov: added `@theia/preview`

lanrat:
> @lanrat @marcdumais-work It sounds like putting projects in GOPATH does not hurt projects using go modules and at the same time provides backward compatibility for projects not using modules. So mounting in GOPATH by default cover all use cases. Do i miss something?

I believe this is the case for Go 1.12. Go 1.11 does not allow modules inside GOPATH.

akosyakov:
https://github.com/golang/go/wiki/Modules#when-do-i-get-old-behavior-vs-new-module-based-behavior

We can allow to pass `GO111MODULE` env variable on docker run that a user can control it. Otherwise i don't know how to satisfy both cases in one image. Would it work for you? @lanrat

marcdumais-work:
I got debug working, thanks @akosyakov. I am ok with merging this PR, pending deciding what to do, if anything, about $GOPATH.

lanrat:
@akosyakov setting the `GO111MODULE` env variable should work just as well.

akosyakov:
I've added README explaining how to control GO modules support and enable debugging: https://github.com/theia-ide/theia-apps/pull/146/files#diff-da05e0fe05bc6624ac724dc91339d747

Will update the desc on Dockerhub as well.

Issue #157, opened 2019-03-22
[java] Update deprecated base image


marcdumais-work:
The `theia-java` [Dockerfile](https://github.com/theia-ide/theia-apps/blob/master/theia-java-docker/Dockerfile#L2) is based on a deprecated image: [ibmcom/ibmnode:8](https://hub.docker.com/r/ibmcom/ibmnode/):

"The ibmcom/ibmnode Docker image has been deprecated following the deprecation of the IBM SDK for Node.js on certain platforms. There will be no further updates to this image. We recommend using the official Node.js Docker image instead, available at https://hub.docker.com/_/node/."

Following that suggestion may potentially also fix this git [issue](https://github.com/theia-ide/theia-apps/issues/156), for `theia-java`.

PR #162, opened 2019-04-09
upgrade next images to use node 10

### LINKED ISSUE #161: [node] Update all docker images to use node 10
Theia is now using [Node 10](https://github.com/theia-ide/theia/pull/4750).

All docker images need to be updated to use this version.

akosyakov:
I've not had time to test them yet. I will do for theia and go images, but i need help with testing other images.

Right now, i don't expect that next images work at all, so help with testing is appreciated!

fix #161

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,20 +11,20 @@ branches:
 language: generic
 
 env:
-  - NPM_TAG=latest IMAGE_NAME=theia-full
-  - NPM_TAG=next IMAGE_NAME=theia-full
-  - NPM_TAG=latest IMAGE_NAME=theia
-  - NPM_TAG=next IMAGE_NAME=theia
-  - NPM_TAG=latest IMAGE_NAME=theia-java
-  - NPM_TAG=next IMAGE_NAME=theia-java
-  - NPM_TAG=latest IMAGE_NAME=theia-go
-  - NPM_TAG=next IMAGE_NAME=theia-go
-  - NPM_TAG=latest IMAGE_NAME=yangster
-  - NPM_TAG=next IMAGE_NAME=yangster
-  - NPM_TAG=latest IMAGE_NAME=theia-python
-  - NPM_TAG=next IMAGE_NAME=theia-python
-  - NPM_TAG=latest IMAGE_NAME=theia-swift
-  - NPM_TAG=next IMAGE_NAME=theia-swift
+  - NPM_TAG=latest IMAGE_NAME=theia-full NODE_VERSION=8.14.1
+  - NPM_TAG=next IMAGE_NAME=theia-full NODE_VERSION=10.15.3
+  - NPM_TAG=latest IMAGE_NAME=theia NODE_VERSION=8
+  - NPM_TAG=next IMAGE_NAME=theia NODE_VERSION=10
+  - NPM_TAG=latest IMAGE_NAME=theia-java NODE_VERSION=8
+  - NPM_TAG=next IMAGE_NAME=theia-java NODE_VERSION=10
+  - NPM_TAG=latest IMAGE_NAME=theia-go NODE_VERSION=8
+  - NPM_TAG=next IMAGE_NAME=theia-go NODE_VERSION=10
+  - NPM_TAG=latest IMAGE_NAME=yangster NODE_VERSION=8
+  - NPM_TAG=next IMAGE_NAME=yangster NODE_VERSION=10
+  - NPM_TAG=latest IMAGE_NAME=theia-python NODE_VERSION=8
+  - NPM_TAG=next IMAGE_NAME=theia-python NODE_VERSION=10
+  - NPM_TAG=latest IMAGE_NAME=theia-swift NODE_VERSION=8
+  - NPM_TAG=next IMAGE_NAME=theia-swift NODE_VERSION=10
 
 install:
   - cd "$IMAGE_NAME-docker"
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -11,20 +11,20 @@ branches:
...
-  - NPM_TAG=next IMAGE_NAME=theia-java
-  - NPM_TAG=latest IMAGE_NAME=theia-go
-  - NPM_TAG=next IMAGE_NAME=theia-go
-  - NPM_TAG=latest IMAGE_NAME=yangster
-  - NPM_TAG=next IMAGE_NAME=yangster
-  - NPM_TAG=latest IMAGE_NAME=theia-python
-  - NPM_TAG=next IMAGE_NAME=theia-python
-  - NPM_TAG=latest IMAGE_NAME=theia-swift
-  - NPM_TAG=next IMAGE_NAME=theia-swift
+  - NPM_TAG=latest IMAGE_NAME=theia-full NODE_VERSION=8.14.1
```
JanKoehnlein: Why do you pin only theia-full to a specific micro version and not the others?

akosyakov: i've just mimicked what was in images before. Should i use latest node for full as well? Not sure what was the reasoning behind.

akosyakov:
@jopit Is it possible to provide `ibmcom/ibmnode:10`? Otherwise we have to exclude java next and latest after releasing 0.6.0.

akosyakov:
The good thing that we have smoke tests. So one can see whether next images are starting up! Awesome, thx @marcdumais-work ❤️

marcdumais-work:
> @jopit Is it possible to provide `ibmcom/ibmnode:10`? Otherwise we have to exclude java next and latest after releasing 0.6.0.

See: https://github.com/theia-ide/theia-apps/issues/157#issue-424216276

akosyakov:
ok, i will try to switch to node images for java

.travis.yml
marcdumais-work: I do not think there is a need for that specific version to be pinned - the various images tend to be updated individually, since some already have Node.js, but others not (i.e. it's hard to be uniform).

akosyakov: I've removed pinning

akosyakov:
@marcdumais-work i've updated java image to use official node image, but i did not test it

.travis.yml
marcdumais-work: Sorry, I was wrong: pinning was correct, since we use that version to fetch a specific tarball from `nodeja.org`, which fails when we only request version `8`. (seems to work for version `10` however)

marcdumais-work: We could however pin to a different version, like the more recent `8.15.1`

marcdumais-work:
BTW I can help with quick testing. I'll start with the full and java.

.travis.yml
```diff
@@ -11,20 +11,20 @@ branches:
...
-  - NPM_TAG=latest IMAGE_NAME=theia-go
-  - NPM_TAG=next IMAGE_NAME=theia-go
-  - NPM_TAG=latest IMAGE_NAME=yangster
-  - NPM_TAG=next IMAGE_NAME=yangster
-  - NPM_TAG=latest IMAGE_NAME=theia-python
-  - NPM_TAG=next IMAGE_NAME=theia-python
-  - NPM_TAG=latest IMAGE_NAME=theia-swift
-  - NPM_TAG=next IMAGE_NAME=theia-swift
+  - NPM_TAG=latest IMAGE_NAME=theia-full NODE_VERSION=8
+  - NPM_TAG=next IMAGE_NAME=theia-full NODE_VERSION=10
```
marcdumais-work: update: it's actually not working for me locally, to use node version "10" for next; we need to pin it to something specific. e.g. using `10.15.3` works, resolving download link for archive to  `https://nodejs.org/dist/v10.15.3/node-v10.15.3-linux-x64.tar.xz`

.travis.yml
```diff
@@ -11,20 +11,20 @@ branches:
...
-  - NPM_TAG=next IMAGE_NAME=theia-java
-  - NPM_TAG=latest IMAGE_NAME=theia-go
-  - NPM_TAG=next IMAGE_NAME=theia-go
-  - NPM_TAG=latest IMAGE_NAME=yangster
-  - NPM_TAG=next IMAGE_NAME=yangster
-  - NPM_TAG=latest IMAGE_NAME=theia-python
-  - NPM_TAG=next IMAGE_NAME=theia-python
-  - NPM_TAG=latest IMAGE_NAME=theia-swift
-  - NPM_TAG=next IMAGE_NAME=theia-swift
+  - NPM_TAG=latest IMAGE_NAME=theia-full NODE_VERSION=8
```
marcdumais-work: Sorry, I was wrong: pinning was correct, since we use that version to fetch a specific tarball from nodejs.org, which fails when we only request version 8.

akosyakov: ok, now we know the reason :D

marcdumais-work:
theia-python and theia-swift also seems to work well (latest and next).

theia-full-docker/Dockerfile
```diff
@@ -30,8 +30,8 @@ RUN set -ex \
     gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys "$key" ; \
     done
 
-
-ENV NODE_VERSION 8.14.1
+ARG NODE_VERSION=10
```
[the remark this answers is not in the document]
akosyakov: fixed!

PR #170, opened 2019-04-24
[electron] add missing @theia/electron dependency

vince-fugnitto:
Added missing `@theia/electron` dependency necessary to build the 
`theia-electron` bundled application.

### DIFF
--- a/theia-electron/package.json
+++ b/theia-electron/package.json
@@ -33,6 +33,7 @@
         "@theia/callhierarchy": "next",
         "@theia/core": "next",
         "@theia/editor": "next",
+        "@theia/electron": "next",
         "@theia/file-search": "next",
         "@theia/filesystem": "next",
         "@theia/git": "next",


marcdumais-work:
Hi @vince-fugnitto ,

I think this is needed because `electron` is no longer a dependency for Theia (switched to a dev-dependency)?  

It looks like only `next` Theia packages are used in this application currently? So no need to wait until `0.6.0` is released to merge this?

vince-fugnitto:
@marcdumais-work 

Yes correct, it is no longer a dependency so in order to build an Electron version we need to explicitly declare it, something that was missing in the current `theia-electron` `package.json`.

As for when it can be merged, I assume that it should be fine to merge now, but if it is a similar case as the following [comment](https://spectrum.chat/theia/general/april-release-v0-6-0~3890ed06-1cb3-46c2-9763-320802a27a80?m=MTU1NjE2ODI5NTM5OQ==), we can always merge after.

marcdumais-work:
I confirmed that this application still builds and the resulting .deb package installs on my Ubuntu laptop, without this fix (master version). However the resulting application does not start - makes sense. 

The version of the application I built from this PR does start however, so I'll approve and merge this right away.

PR #176, opened 2019-05-02
[theia-electron] Use electron-builder js api

marechal-p:
The CLI was marked as soon-to-be deprecated when using
`electron-builder`, so I converted the yaml configuration to a regular
js one.

Also makes the configuration use the electron executable downloaded by
`@theia/electron`, which can be patched to remove proprietary codecs.

Signed-off-by: Paul Maréchal <paul.marechal@ericsson.com>


marechal-p:
cc @kittaakos I think you were the one to setup the Electron bundling?

I stayed as close as possible to what you did, tell me if I messed anything :)

marechal-p:
linked https://github.com/theia-ide/theia/pull/5055

marechal-p:
interesting? https://www.npmjs.com/package/electron-packager-plugin-non-proprietary-codecs-ffmpeg

marcdumais-work:
@marechal-p I have added you as a collaborator on this repo. It doesn't change anything for this specific PR I think (the electron packaging app is not covered by CI), but in the future you can push your PRs branch directly to this repo and have the benefit of working CI.

kittaakos:
> The CLI was marked as soon-to-be deprecated when

Where did you see this?

Can we use pure JS with `// @ts-check` for those two tiny modules?

theia-electron/electron-builder/src/options.ts
```diff
@@ -0,0 +1,70 @@
+import { Configuration } from 'electron-builder';
+
+export function getTheiaBundleConfiguration(preview?: boolean): Configuration {
+    return {
+        electronDist: 'node_modules/electron/dist',
```
kittaakos: Why is this required?

theia-electron/electron-builder/src/options.ts
```diff
@@ -0,0 +1,70 @@
...
+            'package.json'
+        ],
+        win: {
+            target: preview ? 'dir' : [
+                'nsis'
+            ]
+        },
+        mac: {
+            target: preview ? 'dir' : [
+                'dmg',
```
kittaakos: [minor]: `,` but it is up to you.

kittaakos:
I have tried it, it worked 👍 Thank you!

Before approving it; why do we need this change? It's nice we can build it with from JS insead of the CLI but what's the benefit of this change?

theia-electron/electron-builder/src/options.ts
marechal-p: `electronDist` tells `electron-builder` that it shouldn't pull binaries by itself, since we already have Electron installed inside `node_modules` thanks to `@theia/electron`. This distribution is also the one that is being patched by the scripts I added recently, so we really want to take full benefit from it.

marechal-p:
> Before approving it; why do we need this change? It's nice we can build it with from JS insead of the CLI but what's the benefit of this change?

At first I hoped to use some plugin, but I realized it was for a different bundler :(

On the other hand I also have something outputted when packaging on master:

```sh
$ yarn package:preview
yarn run v1.16.0
$ electron-builder --dir
Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.
Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.
  • electron-builder version=20.40.2
[...]
```

marechal-p:
But the two main advantages of this change are:

1. We use the patched Electron binaries pulled by `@theia/electron`.
2. We use the recommended JS API rather than the soon-to-be deprecated CLI one.

kittaakos:
> 1. We use the patched Electron binaries pulled by `@theia/electron`.

OK, so this PR depends on https://github.com/theia-ide/theia/pull/5112, right?

> 2\. We use the recommended JS API rather than the soon-to-be deprecated CLI one.

You might be confused with this: https://github.com/electron-userland/electron-builder/pull/3852

It's not the `electron-builder` CLI that will be deprecated.

kittaakos:
> OK, so this PR depends on [theia-ide/theia#5112](https://github.com/theia-ide/theia/pull/5112), right?

This part works great!

marechal-p:
> [](https://github.com/electron-userland/electron-builder/pull/3852)
> 
> It's not the `electron-builder` CLI that will be deprecated.

Ah indeed, I opened another more minimal change: https://github.com/theia-ide/theia-apps/pull/180

PR #177, opened 2019-05-05
Haskell

stereobooster:


marcdumais-work:
Hi @stereobooster 

FYI no worries about the CI: a Travis security-related limitation does not permit PRs originating from forks to use the GitHub token we setup, resulting in failed CI, most of the time. 

Before we can eventually merge this contribution, you'll need to squash the commits together and sign the resulting commit: `git commit --amend -s`

Please ping me when this is ready for review.

stereobooster:
Current show stopper is https://github.com/theia-ide/haskell-theia-extension/issues/2. I can squash and sign commits 👌

marcdumais-work:
@stereobooster thanks for the info. `@theia/haskell` builds and I get syntax highlighting. We'll try to clarify the current state and get the extension published.

FYI, there might be another way forward: Theia now increasingly is able to re-use VS Code extensions, without modifications. I do not know if any good one exists for Haskell? 

Here's an example of how we modified the golang image, to use the VS Code extension instead of the Theia extension: https://github.com/theia-ide/theia-apps/pull/146

marcdumais-work:
BTW, the current image has quite a massive build time; close to 1.5h on my (not super recent) laptop. Would it be possible to use a pre-compiled stack?

stereobooster:
Stack installs fast, it's building dependencies of HIE and HIE itself takes too long. There is nix-packed version of HIE, not sure if it will be much faster.

marcdumais-work:
@stereobooster About your last commit: any feedback about using the Haskell VS Code extension? Have you had the chance to test it a bit?

stereobooster:
@marcdumais-work not yet it takes infinity to compile. I start the process, but then forgot about it...

marcdumais-work:
@stereobooster ok, I'll also try to build the image locally.

stereobooster:
It took about two hours to build Haskell part.  This is what I get

```
$ download-debug-adapters
vscode-hie-server: downloading from https://ms-vscode.gallery.vsassets.io/_apis/public/gallery/publishers/alanz/vsextensions/vscode-hie-server/0.0.27/vspackage
internal/streams/legacy.js:57
      throw er; // Unhandled stream error in pipe.
```

marcdumais-work:
> **UPD** given url ends with `0.0.27/vspackage` and downloaded file is " application/vsix", which is `tar.gz` file (checked it manually). We need to change logic here https://github.com/theia-ide/theia/blob/master/packages/debug/bin/download-adapters.js#L61
> 
> We can use headers to detect file type:
> 
> * **content-disposition** (`: inline; filename=Microsoft.VisualStudio.Services.VSIXPackage; filename*=utf-8''Microsoft.VisualStudio.Services.VSIXPackage`)
> * **content-type** (`: application/zip; api-version=5.1-preview.1`)

Good find @stereobooster . Would you please open an issue in the [main theia repo](https://github.com/theia-ide/theia/) about this, if none already exists? You are welcome to submit a PR for this too. 

On my side I have confirmed that the VS Code extension here loads without errors in Theia, if manually downloaded first, so it's looking good once we fix the download. 

I have noticed that extension "Haskell Syntax Highlighting" is listed as a dependency; so presumably the syntax highlight is not part of "Haskell Language Server", and so we probably need both. 

image

image

marcdumais-work:
An alternative that does not require a change in the main Theia repo would be to build the extension from [source](https://github.com/alanz/vscode-hie-server) and copy the result into the `plugins` folder in the container.

marcdumais-work:
@stereobooster what's the current status? Does the build succeed for you locally? I'm trying to build now.

stereobooster:
```
Step 12/16 : RUN yarn theia build
 ---> Running in 7bad4c86ee64
yarn run v1.13.0
error Command "theia" not found.
```

stereobooster:
I was able to compile it and run it. But there is no highlighting for Haskell code, no autocomplete for the code, `ghc` is not available in terminal (probably need to provide `PATH`)

marcdumais-work:
> I was able to compile it and run it. But there is no highlighting for Haskell code, no autocomplete for the code, `ghc` is not available in terminal (probably need to provide `PATH`)

These problems may be related. It very possible that the Language Server relies on having `ghc` and maybe other language-specific tools, to be able to function.

dpwiz:
There is a newer [ghcide](https://github.com/digital-asset/ghcide) language server. It's way more lightweight and pretty solid for medium projects.

TheMatten:
One year later again, there's now [Haskell Language Server](https://github.com/haskell/haskell-language-server), which is meant to be final solution for IDE support. Would it make sense to reopen the issue?

vince-fugnitto:
> One year later again, there's now [Haskell Language Server](https://github.com/haskell/haskell-language-server), which is meant to be final solution for IDE support. Would it make sense to reopen the issue?

@TheMatten I do not believe there is a need to support and maintain a `haskell` image since it is essentially similar to other images except that it includes the _haskell language-server_.

marcdumais-work:
There seems to be a [vscode extension that installs the LS if not present](https://github.com/haskell/haskell-language-server#visual-studio-code). We could add this extension to `theia-full`, if it works ok?

Issue #195, opened 2019-07-03
the init process in docker cause a lot of defunct processes


tom-shan:
after start docker with this command:
`
docker run -it -p 3000:3000 -v "$(pwd):/home/project:cached" theiaide/theia:next
`
and refresh the page, there are many zombie processes in the process list.
For now the init process(PID is 0) will not wait for it's child processes, so these child processes will not be reaped by the init and remains.
![image](https://user-images.githubusercontent.com/7686800/60560711-7cb4a180-9d83-11e9-924d-2f00a1c6da69.png)

tom-shan:
According to https://github.com/krallin/tini, we can add a suffix of `"/tini", "--"` to the ENTRYPOINT in Dockerfile.

marcdumais-work:
Interesting, thanks @tom-shan for reporting this issue.

PR #196, opened 2019-07-03
add --init to docker run

tom-shan:
fix https://github.com/theia-ide/theia-apps/issues/195

Signed-off-by: tom-shan <swt0008411@163.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ install:
   - "travis_wait 30 sleep 1800 &"
   - docker build --build-arg "version=$NPM_TAG" --build-arg "NODE_VERSION=$NODE_VERSION" --build-arg "GITHUB_TOKEN=$GITHUB_TOKEN" . -t "$IMAGE_TAG" --no-cache
   - docker tag "$IMAGE_TAG" "$IMAGE:$NPM_TAG"
-  - docker run -d -p 0.0.0.0:4000:3000 "$IMAGE_TAG"
+  - docker run --init -d -p 0.0.0.0:4000:3000 "$IMAGE_TAG"
 
 script: cd ../tests; yarn; yarn test-app-$IMAGE_NAME
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ Building the image for Theia tag:
 
 Running the image:
 
-    docker run -it -p 3000:3000 -v /myproject:/home/project theiaide/theia:v0.2.1
+    docker run -it --init -p 3000:3000 -v /myproject:/home/project theiaide/theia:v0.2.1
 
 Building the next image:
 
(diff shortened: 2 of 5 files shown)


theia-docker/Dockerfile
```diff
@@ -30,5 +30,8 @@ COPY --from=0 --chown=theia:theia /home/theia /home/theia
 EXPOSE 3000
 ENV SHELL /bin/bash
 ENV USE_LOCAL_GIT true
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+RUN chmod +x /tini
```
marcdumais-work: I got a permission issue at first. Got it working with a little tweak, so that the `theia` user has execute permission as well:

`RUN chmod 555 /tini`

theia-docker/Dockerfile
```diff
@@ -30,5 +30,8 @@ COPY --from=0 --chown=theia:theia /home/theia /home/theia
 EXPOSE 3000
 ENV SHELL /bin/bash
 ENV USE_LOCAL_GIT true
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
```
marcdumais-work: I got a runtime exception when running `tini` from the entrypoint. Maybe the basic Ubuntu image is missing some dynamic library that's required? It worked with the statically linked version:

`ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static /tini`

marcdumais-work:
Hi @tom-shan ,

Thanks for this PR. 

Looking at `tini's`documentation, it has been possible for a couple of years to use tiny without changing the Docker images, by using the `--init` command-line option to `docker run`. When doing that, docker injects `tiny` in the image and runs the entrypoint through it (it still works if one added `tini` explicitly in the image). 

Here's an interesting discussion about whether to explicitly use `tini` in docker images or use the docker command-line option: https://github.com/krallin/tini/issues/81 . For me the best argument was that, at the time, a lot of users were likely to run a docker version without this feature. Over two years later, this may not be a big problem anymore. 

So have you considered a documentation update instead of adding `tini` in the image? As it is, we would need to do the same for other images in this repo. If we instead document to use the `--init` option in the README, it would cover all images. 

P.S: Node.js suggests using either `--init` or to include `tini` in images, rather than having node as init process: 
https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#handling-kernel-signals

tom-shan:
I agree with you. Adding `--init` is actually a simpler way.

marcdumais-work:
@tom-shan would you care to submit an updated PR that documents the use of the `--init` docker run option? e.g. in the repo's README.md?

tom-shan:
I just search the entire repo and modify these commands of `docker run`.

marcdumais-work:
> I just search the entire repo and modify these commands of docker run

Up-to you if you think a mention of the reason why `--init` is needed would also be good to add.

tom-shan:
Aha!

PR #198, opened 2019-07-09
RPM build example

jgbradley1:
A proof of concept demonstrating how to build a RPM of the cloud IDE.

### DIFF
--- a/theia-rpm-build-docker/package.json
+++ b/theia-rpm-build-docker/package.json
@@ -1,0 +1,50 @@
+{
+  "name": "theia",
+  "version": "0.0.1",
+  "license": "EPL-2.0",
+  "engines" : {
+	  "node":"10.x"
+  },
+  "dependencies": {
+      "@theia/callhierarchy": "latest",
+      "@theia/file-search": "latest",
+      "@theia/json": "latest",
+      "@theia/markers": "latest",
+      "@theia/merge-conflicts": "latest",
+      "@theia/messages": "latest",
+      "@theia/mini-browser": "latest",
+      "@theia/navigator": "latest",
+      "@theia/outline-view": "latest",
+      "@theia/preferences": "latest",
+      "@theia/preview": "latest",
+      "@theia/search-in-workspace": "latest",
+      "@theia/terminal": "latest",
+      "@theia/textmate-grammars": "latest",
+      "typescript": "latest"
+  },
+  "devDependencies": {
+      "@theia/cli": "latest"
+  },
+  "scripts": {
+      "prebuild": "npm install -g yarn",
+      "prebuild-rpm": "ln -s \"$( pwd )\" ~/rpmbuild && npm install -g speculate",
+      "build-rpm": "speculate && rpmbuild -bb SPECS/theia.spec",
+      "start": "npm run prebuild && yarn theia start"
+  },
+  "spec": {
+    "rebuild": false,
+    "prune": false,
+    "nodeVersion": "> 10.0.0",
+    "requires": [
+      "nodejs",
+      "python"
+    ],
+    "executable": [
+      "./run_theia.sh"
+    ],
+    "post": [
+      "mv /usr/lib/theia/run_theia.sh /usr/local/bin/theia"
+    ]
+  }
+}
+
--- a/theia-rpm-build-docker/Dockerfile
+++ b/theia-rpm-build-docker/Dockerfile
@@ -1,0 +1,34 @@
+FROM centos:7
+# install development tools
+RUN curl --silent --location https://rpm.nodesource.com/setup_10.x | bash - && \
+    yum groupinstall -y "Development Tools" && \
+    yum install -y nodejs && \
+    npm install -g yarn
+
+# required file to install Theia
+COPY package.json run_theia.sh /theia-app/
+
+# build theia
+WORKDIR /theia-app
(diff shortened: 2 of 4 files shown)


marcdumais-work:
Hi @jgbradley1 ,

I think it could be interesting to submit this PR in the [theia-apps](https://github.com/theia-ide/theia-apps) repo? We already have misc Docker images there, as well as one Theia Electron application packaging example.

jgbradley1:
This PR is in the theia-apps repo I believe...or did I mess something up? I'm just having a slight problem signing off on the work so the DCO check passes. Should be fixed soon.

marcdumais-work:
> This PR is in the theia-apps repo I believe...

Confirmed - brainfart on my part I believe :)

theia-rpm-build-docker/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "devDependencies": {
+      "@theia/cli": "latest"
+  },
+  "scripts": {
+      "prebuild": "npm install -g yarn",
+      "prebuild-rpm": "ln -s \"$( pwd )\" ~/rpmbuild && npm install -g speculate",
+      "build-rpm": "speculate && rpmbuild -bb SPECS/theia.spec",
+      "start": "npm run prebuild && yarn theia start"
+  },
+  "spec": {
```
marcdumais-work: I am curious: I do not remember seeing a "spec" section in package.json before. Does the content end-up being used in the packaged Theia application?

marcdumais-work:
I gave this PR a try and it seems to work well. We could tweak the content of the packaged Theia app, but what's there already serves well for a RPM packaging example. I like the added doc too. Good job @jgbradley1 . 

It would be easy to add this app to CI and run the very light smoke test suite on the installed RPM package as well, but until we publish the .rpm, there may not be a strong need. 

The DCO is fixed now. Could you squash the commits? Then once we clarify the "spec" section in `package.json`, I think we're good to go.

theia-rpm-build-docker/package.json
marcdumais-work: I think I got it: it's used by `build-rpm`?

jgbradley1: The `spec` section is used by the `speculate` module. See [here](https://rpm-packaging-guide.github.io/#rpm-packaging-workspace) for what a sample RPM dev workspace looks like. 

The `speculate` module basically generates the entire RPM workspace and all necessary metadata files (including the SPEC file) based on the settings defined in the `spec` section of `package.json`. Once these files are generated, calling the `rpmbuild` utility handles the rest of the packaging process.

jgbradley1:
How would you like to handle the `version` tag in `package.json`? The  `rpmbuild` utility requires that `version` be in proper semantic versioning format. The version info is assigned to the final rpm package and recognized by `yum`.

**Solution 1**: we keep it simple and hardcode `version=1.0.0` as an example. It will not effect the build at all. The version on the final rpm will not align with Theia's current release version though.

**Solution 2**: Track the version number with an environment variable in the Docker image to make the necessary changes to `package.json`, similar to how the env variable `version` is used to differentiate `latest` and `next` builds in the other examples.

marcdumais-work:
> **Solution 1**: we keep it simple and hardcode `version=1.0.0` as an example. It will not effect the build at all. The version on the final rpm will not align with Theia's current release version though.

I think an hardcoded version will do fine, for such an example, where the value lies more in the demonstration of how to nicely package a Theia app as a RPM vs making a Theia application for specific language(s)/purpose(s).

jgbradley1:
I'll change the files so `version=0.0.1`, so it also aligns with your suggestion in the [Debian build example](https://github.com/theia-ide/theia-apps/pull/199).

jgbradley1:
Commits are squashed. I have not previously squashed commits so let me know if something is off.

theia-rpm-build-docker/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "rebuild": false,
+    "prune": false,
+    "nodeVersion": "> 10.0.0",
+    "requires": [
+      "nodejs",
+      "python"
```
marcdumais-work: I noticed that `git` is currently missing, which causes exceptions at Theia startup. If we add it here, will it then be installed along with the Theia RPM? The version may be a bit old vs what Theia is expecting, however. 

Removing extension `@theia/git` would also fix the issue.

marcdumais-work: BTW, general question: here CentOS 7 is used to generate and install/test the RPM. Would the RPM be installable on later versions of the OS as well?

jgbradley1: I removed `@theia/git` extension for now to keep things simple.

Regarding using a later version of the OS, according to the [speculate](https://github.com/bbc/speculate) docs, it currently supports CentOS 7. I made an [issue](https://github.com/bbc/speculate/issues/75) asking if they plan to support CentOS 8. The module is backed by BBC news so it probably has more support than a normal open source project.

marcdumais-work:
@jgbradley1 Can you add a significant commit message, describing the content of this contribution? Does not need to be long or anything. I would mention the current limitation (IIUC) that the generated RPM would work on CentOS/RH Linux 7.

Otherwise looking good!

jgbradley1:
Done. Let me know if you would like more info added to the commit message.

PR #199, opened 2019-07-10
Debian build example

jgbradley1:
This is a proof of concept demonstrating how to wrap Theia up into a debian package.

Signed-off-by: Joshua Bradley <jgbradley1@gmail.com>

### DIFF
--- a/theia-deb-build-docker/Dockerfile
+++ b/theia-deb-build-docker/Dockerfile
@@ -1,0 +1,37 @@
+FROM ubuntu:bionic
+# install development tools
+RUN apt-get update && \
+    apt-get install -y curl sudo build-essential jq vim && \
+    curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash - && \
+    apt-get install -y nodejs && \
+    npm install -g yarn
+
+# add theia installation file
+COPY package.json run_theia.sh /theia-app/
+RUN chmod +x /theia-app/run_theia.sh
+
+# build theia
+WORKDIR /theia-app
+RUN yarn && \
+    NODE_OPTIONS="--max_old_space_size=4096" yarn theia build && \
+    yarn autoclean --init && \
+    echo *.ts >> .yarnclean && \
+    echo *.ts.map >> .yarnclean && \
+    echo *.spec.* >> .yarnclean && \
+    yarn autoclean --force && \
+    yarn cache clean
+
+# create debian package
+RUN npm run build-deb
+
+# transfer debian package to a new clean image and install theia
+FROM ubuntu:bionic
+COPY --from=0 /theia-app/theia-example_0.0.1_all.deb /
+RUN apt-get update && apt-get install -y curl sudo && \
+    curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash - && \
+    apt-get install -y ./theia-example_0.0.1_all.deb && \
+    rm theia-example_0.0.1_all.deb
+
+ENV SHELL /bin/bash
+EXPOSE 3000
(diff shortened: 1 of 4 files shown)


theia-deb-build-docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+COPY package.json run_theia.sh /theia-app/
+RUN chmod +x /theia-app/run_theia.sh
+
+# build theia
+WORKDIR /theia-app
+RUN yarn && \
+    NODE_OPTIONS="--max_old_space_size=4096" yarn theia build && \
+    yarn autoclean --init && \
+    echo *.ts >> .yarnclean && \
+    echo *.ts.map >> .yarnclean && \
```
marcdumais-work: I guess it would be an option to package the `.ts` and `.ts.map` files separately in their own .deb package, instead of deleting them? i.e. a bit like for some C apps, one can sometimes install the `<app>-debug` package, if the debug symbols are needed.
(does not need to be done as part of this PR necessarily)

theia-deb-build-docker/package.json
```diff
@@ -0,0 +1,43 @@
+{
+    "name": "theia",
+    "version": "0.8.0",
```
marcdumais-work: I would suggest having a version for this demo app that's distinct from the Theia platform version. Else it will need to be updated often or will not reflect reality. Specially since this is then used to name the .deb package, and so is more visible. Suggestion 0.0.1 :) The name as well could be something like "debian-packaging-example" or something similar.

theia-deb-build-docker/Dockerfile
jgbradley1: Certainly. I only remove those files because the other examples in the repo do it. I wanted to produce the smallest deb package possible. I don't know if the [node-deb](https://www.npmjs.com/package/node-deb) module supports creating a debug package explicitly. It has good docs and lots of examples but makes no reference to building debug packages.

I suppose one could leave the source files alone (not remove them) and then continue the build process in order to create an `<app>-debug` package.

theia-deb-build-docker/package.json
jgbradley1: The version is now hardcoded to be `version=0.0.1`. The node-deb module only uses the `name` field in `package.json` to name the final `.deb` file.

For example, if `name=theia-test` and `version=0.0.1`, the final .deb file would be called `theia-test_0.0.1_all.deb`. The filename does not affect calling `theia` from the commandline once the package has been installed by apt.

jgbradley1:
I removed `@theia/git` from `package.json` for the sake of simplicity. If someone did want to use the git extension, they would add `git` to the `dependencies` under the `node-deb` section just like I do with `nodejs` in order for apt to properly install everything.

`node-deb` has decent documentation showing how to include additional dependencies which is why I provide a direct link to the module docs in the README.

theia-deb-build-docker/package.json
marcdumais-work: Thanks for the explanation. 

How about using `theia-example` as name, then? It's a minor detail, but we can see that there is already quite a bit of confusion between `Theia` the platform and calling specific applications the same.  e.g. think of calling an IDE developed on the Eclipse platform "eclipse"  ;)

marcdumais-work:
@jgbradley1 Can you set a descriptive commit title/message on this one as well?

theia-deb-build-docker/package.json
jgbradley1: Changed `name=theia-example`.

marcdumais-work:
Tested locally - works as documented and can serve as a nice example of a Debian packaging for a Theia app.

PR #200, opened 2019-07-10
Adds rust support

dwjbosman:
This branch adds a theia rust Docker image.


marcdumais-work:
Hi @dwjbosman FYI, the Travis CI failures are expected when a PR for this repo originates from a fork, so no worries. 

Preliminary question: this seems like a substantial contribution and I do not see any copyright or license information within: is this all new, original code/content, that you have permission to submit under this repo's Apache 2 license?

dwjbosman:
1. All code/scripts included in the fork itself are written by me and can be published under Apache 2.0. When the docker image builds it will include various third party libraries (eg. theia rust extension, theia cpp). Most of these packages also have Apache 2.0 licenses, Theia-cpp-extension has a Eclipse 2.0 license.
2. I haven't included any travis testing yet. I will propose a test step that builds the image.

dwjbosman:
I have been trying to get travis working. Unfortunately my travis build fails on building my Docker image: "yarn theia build" fails with  "JavaScript heap out of memory".  Now travis workers have only 3GB of memory. Have you experience out of memory errors on any of the other Theia images?

dwjbosman:
I have setup Travis to use another version of Ubuntu (Xenial). Now I am getting this error: 

"Downloading ripgrep failed: Error: Request failed with code 401"

This seems a problem other users experience as well. @marcdumais-work have you ever experienced this error?

dwjbosman:
Fixed
"Downloading ripgrep failed: Error: Request failed with code 401""
by adding a GIT_TOKEN. Now I'm getting the "JavaScript heap out of memory" again. I ran "free' just before starting theia build.

```
Step 14/29 : RUN free -h
 ---> Running in bdb8563561f0
              total        used        free      shared  buff/cache   available
Mem:           7.3G        376M        4.3G        8.6M        2.6G        6.6G
Swap:            0B          0B          0B
```
This doesn't look too bad.

I am no going to try a node option: --max-old-space-size=8192

dwjbosman:
Memory problems solved. Now travis fails on the "yarn test" command. What does the test application expect from my Docker image?

.travis.yml
```diff
@@ -11,20 +13,21 @@ branches:
...
+#  - NPM_TAG=next IMAGE_NAME=theia-java NODE_VERSION=10
+#  - NPM_TAG=latest IMAGE_NAME=theia-go NODE_VERSION=10
+#  - NPM_TAG=next IMAGE_NAME=theia-go NODE_VERSION=10
+#  - NPM_TAG=latest IMAGE_NAME=yangster NODE_VERSION=10
+#  - NPM_TAG=next IMAGE_NAME=yangster NODE_VERSION=10
+#  - NPM_TAG=latest IMAGE_NAME=theia-python NODE_VERSION=10
+#  - NPM_TAG=next IMAGE_NAME=theia-python NODE_VERSION=10
+#  - NPM_TAG=latest IMAGE_NAME=theia-swift NODE_VERSION=10
+#  - NPM_TAG=next IMAGE_NAME=theia-swift NODE_VERSION=10
+   - NPM_TAG=next IMAGE_NAME=theia-rust NODE_VERSION=Dummy
```
vince-fugnitto: Is there a reason for having `NODE_VERSION=Dummy`?

.travis.yml
```diff
@@ -11,20 +13,21 @@ branches:
...
-  - NPM_TAG=next IMAGE_NAME=theia-java NODE_VERSION=10
-  - NPM_TAG=latest IMAGE_NAME=theia-go NODE_VERSION=10
-  - NPM_TAG=next IMAGE_NAME=theia-go NODE_VERSION=10
-  - NPM_TAG=latest IMAGE_NAME=yangster NODE_VERSION=10
-  - NPM_TAG=next IMAGE_NAME=yangster NODE_VERSION=10
-  - NPM_TAG=latest IMAGE_NAME=theia-python NODE_VERSION=10
-  - NPM_TAG=next IMAGE_NAME=theia-python NODE_VERSION=10
-  - NPM_TAG=latest IMAGE_NAME=theia-swift NODE_VERSION=10
-  - NPM_TAG=next IMAGE_NAME=theia-swift NODE_VERSION=10
+#  - NPM_TAG=latest IMAGE_NAME=theia-full NODE_VERSION=10.15.3
```
vince-fugnitto: Is there a reason for commenting these out?

.travis.yml
```diff
@@ -35,13 +38,15 @@ install:
   # from https://stackoverflow.com/questions/43918874/how-to-increase-no-activity-wait-time-in-travis-ci
   - "travis_wait 30 sleep 1800 &"
   - docker build --build-arg "version=$NPM_TAG" --build-arg "NODE_VERSION=$NODE_VERSION" --build-arg "GITHUB_TOKEN=$GITHUB_TOKEN" . -t "$IMAGE_TAG" --no-cache
+#  - docker build --build-arg "version=$NPM_TAG" --build-arg "NODE_VERSION=$NODE_VERSION" --build-arg "GITHUB_TOKEN=$GITHUB_TOKEN" --build-arg "NODE_OPTIONS=$NODE_OPTIONS" . -t "$IMAGE_TAG" --no-cache
```
vince-fugnitto: I'm not sure commenting out this command is appropriate, it means that other images cannot use the 
command any longer.

dwjbosman: The change is actually an addition by me to be able to pass NODE_OPTIONS via travis settings to the container. This was originally not possible. This is not needed anymore and I will rollback this modification.

.travis.yml
dwjbosman: This temporary. I wanted to test my pull request with Travis, without running the tests for the other images.

.travis.yml
dwjbosman: My image is directly based on Ubuntu and an installation of yarn using apt-get. This automatically installs node 8.  I will look into "nvm" to see if I can install the node version + yarn requested by NODE_VERSION

dwjbosman: I updated it to use nvm. It nos respects NODE_VERSION

.travis.yml
dwjbosman: The change was reverted. During testing by Travis I am getting failed tests:

`[05:25:34]  	Saved screenshot: ERROR_chrome_2019-07-16T05-25-34.888Z.png`

How do I access these images?

marcdumais-work:
> I have setup Travis to use another version of Ubuntu (Xenial). Now I am getting this error:
> 
> "Downloading ripgrep failed: Error: Request failed with code 401"
> 
> This seems a problem other users experience as well. @marcdumais-work have you ever experienced this error?

Yes, this is a combination of three things: 
- we use `vscode-ripgrep` as a production dependency, and this NPM package downloads a platform-specific binary of significant size, from GitHub. 
- GitHub has a protection mechanism to prevent abuse: an "API rate limit". It looks like Travis executor machines often use resources from GitHub, and so trigger that limit. When that limit is reached, file downloads are denied.
- When one is authenticated with GH, the limit is raised. So as part of our Travis job, we use a GH token so the job is run "authenticated". For security reasons, Travis only uses the token for PRs that originate from this repo, not forks. 

In consequence, non-committers to this repo that need to use their fork to submit PRs will often hit the "API rate limit" of GH, and so have their CI fail, since the Theia application in their image can't build. 

We have unfortunately not found a way to prevent this. As a mitigation, I usually push the PR on my personal fork and have it build there, before merging, to make sure it builds.

.travis.yml
marcdumais-work: > How do I access these images?

Not sure if it's possible, when running on Travis. However the `wdio` test suite we run ATM is very superficial, and mostly tests that the Theia application is able to start, serve a front-end and has the basic expected workbench layout/views. 

You can try to run it locally and then you'd be able to access the screenshot. Quickly: 
- build an image, start it using external port `4000`: `docker run --init -d -p 0.0.0.0:4000:3000 <$IMAGE_TAG>`
- `cd <repo root>/tests`
- `yarn; yarn test-app-$IMAGE_NAME`  # where $IMAGE_NAME = e.g. `theia-java`, `theia-full`, ...

See file `.travis.yml` in the repo's root for more details

marcdumais-work:
> Memory problems solved. Now travis fails on the "yarn test" command. What does the test application expect from my Docker image?

Basically that a Theia application be listening on `localhost`, port `4000`, and that the initial workbench layout of the Theia application be what's usually expected.

marcdumais-work:
Hi @dwjbosman ,

I built the docker image from this PR and can confirm that the tests fail locally as well. Looks like the Theia application does not respond and the tests end-up with timeout. 

However if I start the container by overriding the entrypoint and then start the Theia application manually in there, then it works fine for the tests as well as connecting to it using a web browser (localhost:4000)

```bash
$> docker run --entrypoint bash -it -p 4000:3000 theia-rust:next
root@365ab96e99f6:~/theia_rust_extension# yarn theia start /home/project --hostname=0.0.0.0
```

marcdumais-work:
@dwjbosman FYI - not sure which OS you use to work, but if it happens to be Ubuntu 18.04, you may ATM have problems with building the test suite in this repo. 

If so this PR should help: https://github.com/theia-ide/theia-apps/pull/203

theia-rust-docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+EXPOSE 3000
+ENV SHELL /bin/bash
+# script to start theia
+ADD run.sh /root/run.sh
+# script to set environment
+ADD bashrc /root/bashrc
+RUN chmod 755 /root \
+    && chmod 755 /root/bashrc
+WORKDIR /root/theia_rust_extension
+CMD [ "yarn", "theia", "start", "/home/project", "--hostname=0.0.0.0" ]
```
marcdumais-work: I was able to make it work by using this entrypoint, specifying the absolute path for the node executable:
`ENTRYPOINT [ "/root/.nvm/versions/node/v10.16.0/bin/node", "/root/theia_rust_extension/src-gen/backend/main.js", "/home/project", "--hostname=0.0.0.0" ]`

It's a bit fragile however, since the node path will change with the node version. 

Without the full path, node is not found and so Theia can't start. I  think maybe the `nvm` configuration is not active when the entrypoint is executed. If I instead override the entrypoint with `bash`, then node is in the path, presumably because root's `.bashrc` is read and configures `.nvm`.

dwjbosman:
Dear Marc,
Thank you for the debugging. I am working on a robust solution.
regards,Dinne

    On Tuesday, July 16, 2019, 7:14:49 PM GMT+2, Marc Dumais <notifications@github.com> wrote:  
 
 
@marcdumais-work commented on this pull request.

In theia-rust-docker/Dockerfile:
> +RUN . /root/rust/cargo/env && rustup default $rust_channel
+RUN . /root/rust/cargo/env && rustup component add rust-src rust-docs rls rust-analysis rustfmt clippy
+
+
+RUN . /root/rust/cargo/env && cargo +$rust_channel install racer
+
+EXPOSE 3000
+ENV SHELL /bin/bash
+# script to start theia
+ADD run.sh /root/run.sh
+# script to set environment
+ADD bashrc /root/bashrc
+RUN chmod 755 /root \
+    && chmod 755 /root/bashrc
+WORKDIR /root/theia_rust_extension
+CMD [ "yarn", "theia", "start", "/home/project", "--hostname=0.0.0.0" ]

I was able to make it work by using this entrypoint, specifying the absolute path for the node executable:
ENTRYPOINT [ "/root/.nvm/versions/node/v10.16.0/bin/node", "/root/theia_rust_extension/src-gen/backend/main.js", "/home/project", "--hostname=0.0.0.0" ]

It's a bit fragile however, since the node path will change with the node version.

Without the full path, node is not found and so Theia can't start. I think maybe the nvm configuration is not active when the entrypoint is executed. If I instead override the entrypoint with bash, then node is in the path, presumably because root's .bashrc is read and configures .nvm.

theia-rust-docker/Dockerfile
dwjbosman: I fixed the issue by calling nvm_setup as part of /root/bashrc. The tests succeed now. 

I also added a commit_screenshots.sh script. This scipt is called when a test failure occurs. It will commit all screenshot pngs to a 'bugs' branch. This branch is emptied at every push.  If you want to use this script a separate empty branch has to be created to which travis can commit the screenshots.

Please let me know if you need any other rework.

.travis.yml
dwjbosman: Reverted the changes

.travis.yml
dwjbosman: Ok. See later comment on addition of a "commit_screenshots.sh" script.

marcdumais-work:
@dwjbosman thanks for the updates - I will have another look. 

In the meantime, I think there was a rebase issue at some point: I see in your PR a mix of your own commits and also some that are already merged into master. I hope the net effect is nil for the existing commits, and if so squashing all PRs commits into one should take care of that. So can you squash everything down to one commit, ideally, unless you have a different breakdown in mind that makes sense.

.travis.yml
```diff
@@ -42,8 +43,12 @@ install:
 
 script: cd ../tests; yarn; yarn test-app-$IMAGE_NAME
 
+after_failure:
+  script: cd $TRAVIS_BUILD_DIR && ./commit_screenshots.sh
```
marcdumais-work: So, if I understand correctly, to take advantage of `commit_screenshots.sh`, we would need to create a "bugs" branch in this repo, and then the script will be able to commit any error screenshot generated during CI to that branch?

commit_screenshots.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
marcdumais-work: Can you please add a copyright header for this file? The copyright owner would be yourself or your employer, depending on your specific situation. The license should be the same as in the repo (Apache 2.0)

.travis.yml
dwjbosman: Correct

tests/package.json
```diff
@@ -3,15 +3,16 @@
   "name": "tests_for_theia-apps",
   "version": "0.0.1",
   "scripts": {
-    "test": "wdio wdio.conf.js --theia-port 4000",
-    "test-app-theia": "wdio wdio.theia-docker.conf.js --theia-port 4000",
+    "test": "wdio wdio.conf.js --theia-port 3000",
```
marcdumais-work: When we run CI on Travis, running the Theia apps on whatever port we chose would be ok. 

However we originally chose 4000 because when running tests locally, Theia contributors will often have many instances running on their machine, and a collision is likely when using default port 3000 (and 3001, 3002,...). 

Can you roll-back to port 4000?

theia-rust-docker/next.package.json
```diff
@@ -0,0 +1,51 @@
...
+        "@theia/process": "next",
+        "@theia/rust": "next",
+        "@theia/search-in-workspace": "next",
+        "@theia/task": "next",
+        "@theia/terminal": "next",
+        "@theia/textmate-grammars": "next",
+        "@theia/userstorage": "next",
+        "@theia/variable-resolver": "next",
+        "@theia/workspace": "next",
+        "theia-yang-extension": "next",
```
marcdumais-work: Is this extension needed for your purposes? It's to support the [YANG language](https://en.wikipedia.org/wiki/YANG), which is rather niche.

theia-rust-docker/next.package.json
```diff
@@ -0,0 +1,51 @@
...
+        "@theia/merge-conflicts": "next",
+        "@theia/messages": "next",
+        "@theia/metrics": "next",
+        "@theia/mini-browser": "next",
+        "@theia/monaco": "next",
+        "@theia/navigator": "next",
+        "@theia/outline-view": "next",
+        "@theia/output": "next",
+        "@theia/plantuml": "next",
+        "@theia/plugin": "next",
```
marcdumais-work: If you do not require compatibility with VS Code extensions, you could save space and build time by removing `@theia/plugin*`

theia-rust-docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+ADD install_node.sh /root/install_node.sh
+RUN bash /root/install_node.sh --version $NODE_VERSION
+ADD nvm_setup.sh /root/nvm_setup.sh
+
+#install theia
+RUN mkdir /root/theia_rust_extension
+RUN . /root/nvm_setup.sh && cd /root/theia_rust_extension && yarn global add node-gyp
+ADD next.package.json /root/theia_rust_extension/package.json
+RUN git clone https://github.com/eclipse/theia-cpp-extension.git /root/theia-cpp-extension
+RUN cp -R /root/theia-cpp-extension/packages /root/theia_rust_extension/.
```
marcdumais-work: nitpick: `/root/theia_rust_extension` looks like the place where the Theia rust application is build, and so may be better named something like `theia_rust_app` or similar. Could be defined as a ARG or ENV.

theia-rust-docker/bashrc
```diff
@@ -0,0 +1,19 @@
...
+    export CONTAINER_START_PATH="/root"
+fi  
+
+if [ -z "$HOME" ]; then
+    export HOME="/root"
+fi
+
+echo "HOME: $HOME"
+echo "CONTAINER_START_PATH: $CONTAINER_START_PATH"
+cd "$CONTAINER_START_PATH"
```
marcdumais-work: Is there an implicit assumption that CONTAINER_START_PATH be under the user's home directory, and so mounted in the container? Mine is not, and so when I execute `ex_bash`, I get at the end: 

```bash
CONTAINER_START_PATH: /opt/ericsson-dev/theia-apps/theia-rust-docker
bash: cd: /opt/ericsson-dev/theia-apps/theia-rust-docker: No such file or directory
```
If so it can be an acceptable limitation, but maybe add a note about it in `README.md`?

theia-rust-docker/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Start Theia
+
+1. cd to a folder in your home dir.
+
+eg. cd to the example\_workspace folder.
+  
+2. Start editor
+
+<pre>[path to theia_apps/theia-rust-docket]/ex_theia</pre>
```
marcdumais-work: typo, I think: docket -> docker? Same repeats a few time in this file

theia-rust-docker/README.md
```diff
@@ -0,0 +1,41 @@
+# Theia with Rust support
+
+## Building the image
+
+RUST\_CHANNEL=nightly make image
```
marcdumais-work: Would be good to add a bit of context: it's assumed that one is in the `theia-rust-docker` folder when running this command, I think? 

e.g.: 
```
cd <this repo>/theia-rust-docker
RUST_CHANNEL=nightly make image
```

theia-rust-docker/README.md
marcdumais-work: I thought that the format was a bit confusing here and in other similar occurrences in this file. Maybe the following would be clearer? 
`[path to theia-apps]/theia-rust-docker/ex_theia`

marcdumais-work: Quickly tested `ex_theia`: worked well

theia-rust-docker/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+<pre>[path to theia_apps/theia-rust-docket]/ex_theia</pre>
+
+This creates a container and mounts your home dir. It creates a user inside the container with matching UID/GID so that you can edit files without corrupting permissions. A volume rust-cargo-vol is created in which the rust environment is stored. This will allow you to persistently install crates etc. across container life cycles.
+
+## Bash inside the container
+
+3. Start a bash within the container
+
+<pre>[path to theia_apps/theia-rust-docket]/ex_bash</pre>
```
marcdumais-work: Quickly tested `ex_bash`: worked well

theia-rust-docker/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+<pre>[path to theia_apps/theia-rust-docket]/ex "...command..."</pre>
+
+eg.
+
+<pre>ex "cargo --list"</pre>
+
+## Example project
+
+A very simple rust example project is included in the example_project folder. It contains a build task and debug configuration which uses rust-gdb.
```
marcdumais-work: Quickly tested debugging with the example project: worked well!

theia-rust-docker/install_node.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/sh
+
+# Install node and npm via nvm - https://github.com/creationix/nvm
```
marcdumais-work: There is a little glitch when starting a Theia terminal that seems related to using `nvm`. The following printout is output each time:
`nvm is not compatible with the npm config "prefix" option: currently set to "/root/.nvm/versions/node/v10.16.0"
Run `npm config delete prefix` or `nvm use --delete-prefix v10.16.0 --silent` to unset it.`

theia-rust-docker/next.package.json
```diff
@@ -0,0 +1,51 @@
+{
+    "private": true,
+    "dependencies": {
+        "@theia/callhierarchy": "next",
```
marcdumais-work: consider: This extension could be removed - this feature does not seem to be supported by the rust Language Server and if kept there will be an entry for it in the `view` menu.

theia-rust-docker/next.package.json
```diff
@@ -0,0 +1,51 @@
...
+        "@theia/search-in-workspace": "next",
+        "@theia/task": "next",
+        "@theia/terminal": "next",
+        "@theia/textmate-grammars": "next",
+        "@theia/userstorage": "next",
+        "@theia/variable-resolver": "next",
+        "@theia/workspace": "next",
+        "theia-yang-extension": "next",
+
+        "typescript": "latest"
```
marcdumais-work: Probably not useful for this image

theia-rust-docker/next.package.json
```diff
@@ -0,0 +1,51 @@
+{
+    "private": true,
```
marcdumais-work: Consider adding a name for this Theia application. See e.g.: [here](https://github.com/theia-ide/theia-apps/pull/204/files#diff-5dc95d46717c4be6967fb46980c2f064R3). 

I am not sure if the way the user's home directory is mounted in the image if it might make sense to enable the "thrash" can for deleted files? If so you can omit this preference since the default is enabled (true).

theia-rust-docker/Dockerfile
marcdumais-work: > I fixed the issue by calling nvm_setup as part of /root/bashrc. The tests succeed now.

Confirmed: now works locally for me

marcdumais-work:
@dwjbosman Having now completed a full pass, I realize that there is a lot more here than just our usual `Dockerfile` and `package.json`. This is closer to a nice "solution" than a mere demo; thanks for pushing the limits :)

commit_screenshots.sh
dwjbosman: Added

tests/package.json
dwjbosman: Reverted

theia-rust-docker/next.package.json
dwjbosman: Removed

theia-rust-docker/next.package.json
dwjbosman: Is required. The cpp-debugging plugin depends on some vs code.

theia-rust-docker/Dockerfile
dwjbosman: Created an ENV var

theia-rust-docker/bashrc
dwjbosman: Changed readme a bit. Cleaned up the ex command script. Will only change dir if it is possible.

theia-rust-docker/README.md
dwjbosman: Changed

theia-rust-docker/README.md
dwjbosman: Updated it.

theia-rust-docker/README.md
dwjbosman: Cleaned up the script

theia-rust-docker/README.md
dwjbosman: Cleaned up the script

theia-rust-docker/install_node.sh
dwjbosman: Will look into this

theia-rust-docker/next.package.json
dwjbosman: Removed

theia-rust-docker/next.package.json
dwjbosman: Removed

theia-rust-docker/next.package.json
dwjbosman: Added

akosyakov:
Just fyi Theia is capable to run VS Code `rust` extension. It should be better than native Theia extension.

dwjbosman:
> Just fyi Theia is capable to run VS Code `rust` extension. It should be better than native Theia extension.

Ok, I'll have a look at it. Maybe it could be an configurable option to choose either for the vscode or native plugin. 
@akosyakov Could you point out how package.json should be modified so that it uses the vscode plugins?

marcdumais-work:
>  Could you point out how package.json should be modified so that it uses the vscode plugins?

Here's an example: 
https://github.com/theia-ide/theia-apps/blob/master/theia-go-docker/next.package.json#L19-L21

theia-rust-docker/next.package.json
marcdumais-work: I don't think so: @theia/cpp-debug is purely a Theia extension. Maybe the presence of a `.vscode` folder in the repo made it seem that way, but it just contains some repo-specific IDE configs that Theia can also consume, at least in good part. (in the early days of Theia, most of us were using VS Code to develop Theia)

In any case it's ok to keep the plugin-related extensions, specially if we'd like to try using the VS Code rust extension instead of the corresponding Theia extension.

theia-rust-docker/README.md
```diff
@@ -0,0 +1,44 @@
+# Theia with Rust support
+
+## Building the image
+
+<pre>
+cd <this repo>/theia-rust-docker
```
marcdumais-work: I think you need to escape the `<` and `>` characters in MarkDown. ATM the rendering looks like:
image

dwjbosman:
> > Could you point out how package.json should be modified so that it uses the vscode plugins?
> 
> Here's an example:
> https://github.com/theia-ide/theia-apps/blob/master/theia-go-docker/next.package.json#L19-L21

So I am thinking about creating two images, one based on vscode-rust and the other on the native theia rust extension. 

Currently the way the tests are organised only allows for one test per directory/image. But now I want to run 2 tests (rust native theia and ruse vscode). I think there are two options:
1. Add another directory: theia-native-rust-docker. Copy everything and modify next.package.json

2. 
2.1. Create a symlink theia-native-rust-docker to theia-rust-docker. Add the symlink to git.
2.2. Modify .travis.yml:

```
env:
  - NPM_TAG=latest IMAGE_NAME=theia-native-rust NODE_VERSION=10.15.3 BUILD_FLAG="--build-arg NATIVE"
.....
docker build $BUILD_FLAG --build-arg "version=$NPM_TAG" --build-arg "NODE_VERSION=$NODE_VERSION" --build-arg "GITHUB_TOKEN=$GITHUB_TOKEN" . -t "$IMAGE_TAG" --no-cache
```

My preference goes to option 2. Still it's not very nice.  Do you have any idea how to setup more tests for a single image folder?

marcdumais-work:
Hi @dwjbosman 

> So I am thinking about creating two images, one based on vscode-rust and the other on the native theia rust extension.

Maybe we should keep things simple and chose one or the other? I think that if the Theia native rust extension breaks, we would be tempted to just try the one from VS Code and switch, assuming it works well enough. 

Do you have other reasons to want both?

marcdumais-work:
@dwjbosman There is already quite a bit of content in this PR, so I think it would be fine to continue to use the Theia rust extension for now, and consider switching to the VS Code rust extension later or e.g. in a follow-up PR, by you or someone else.

theia-rust-docker/README.md
dwjbosman: Updated

theia-rust-docker/install_node.sh
dwjbosman: Fixed

dwjbosman:
I squashed all my commits in one commit and signed it off. Due to earlier issues I had a hard time getting this right. Somehow github determines that I have changed 91 files.....

marcdumais-work:
> I squashed all my commits in one commit and signed it off. Due to earlier issues I had a hard time getting this right. Somehow github now determines that I have changed 91 files.....

Hi @dwjbosman . I think there was probably a rebase mishap at some point - the commits added to the PR along with yours are already in the repo, but under different SHAs. Here's a way I think you can fix this: 

```
# start fresh from this repo's master and cherry-pick your squashed commit on top 
# (assuming this repo's remote is name "upstream" and your fork's remote is "origin"; 
# adjust if necessary)
git checkout master
git fetch upstream master
git reset --hard upstream master
git cherry-pick 057c03bba2c6ad64dcb6768728dc2fe000e742c4
git push -f origin
```
I have done similar as above and pushed the result on my own fork (I do not have write permission to push to yours):
https://github.com/theia-ide/theia-apps/pull/213

Please let me know if you need help with this.

dwjbosman:
> > I squashed all my commits in one commit and signed it off. Due to earlier issues I had a hard time getting this right. Somehow github now determines that I have changed 91 files.....
> 
> Hi @dwjbosman . I think there was probably a rebase mishap at some point - the commits added to the PR along with yours are already in the repo, but under different SHAs. Here's a way I think you can fix this:
> 
> ```
> # start fresh from this repo's master and cherry-pick your squashed commit on top 
> # (assuming this repo's remote is name "upstream" and your fork's remote is "origin"; 
> # adjust if necessary)
> git checkout master
> git fetch upstream master
> git reset --hard upstream master
> git cherry-pick 057c03bba2c6ad64dcb6768728dc2fe000e742c4
> git push -f origin
> ```
> 
> I have done similar as above and pushed the result on ~my own fork~ here (I do not have write permission to push to yours):
> #213
> 
> Please let me know if you need help with this.

Thanks. I performed the reset and cherry picking. This looks ok now.

marcdumais-work:
> Thanks. I performed the reset and cherry picking. This looks ok now.

I'll test the result quickly to make sure nothing was broken during the git operations.

theia-rust-docker/README.md
```diff
@@ -0,0 +1,44 @@
+# Theia with Rust support
+
+## Building the image
+
+<pre>
+cd [this repo]/theia-rust-docker
+RUST_CHANNEL=nightly make image
```
marcdumais-work: The nightly channel does not seem to build right now. Add an example here with stable channel or without mentioning channel so that the known-to-work default in Dockerfile is used?

marcdumais-work: As well, since make target "image" does not use "--no-cache", add an example with "make clean-image" ?

marcdumais-work:
@dwjbosman It seems that I am no longer able to successfully start a container based on the image unless I use the provided scripts. 

e.g. if I build the image and attempt to run it like so, the container exits immediately: 
```
docker build --no-cache  -t theia-rust:next .
docker run  -it -p 3000:3000 theia-rust:next
```

This is similar to how we build/test in CI, so I expect that it will fail there as well. 

After a bit of debugging, I think that the way `.bashrc` is sourced from `bashrc` is bogus: it does not correctly set the environment, when starting the container like above. Subsequently. when `run.sh` is executed, node is not in the path, and so the Theia app can't start. You can confirm by adding `echo "PATH: $PATH"` in `run.sh`

If instead one starts the container with command `bash`, then `.bashrc` is read and the environment is set correctly, and then one can start Theia. 

Can you attempt to fix this? 

BTW, I think this type of issue could be avoided if something else than `nvm` was used to pull node.js (e.g. install from archive and manually set PATH from image) ; then there would be no requirement for .bashrc to be read before being able to start Theia - $PATH as set in the image would be enough.

dwjbosman:
> @dwjbosman It seems that I am no longer able to successfully start a container based on the image unless I use the provided scripts.
> 
> e.g. if I build the image and attempt to run it like so, the container exits immediately:
> 
> ```
> docker build --no-cache  -t theia-rust:next .
> docker run  -it -p 3000:3000 theia-rust:next
> ```
> 
> This is similar to how we build/test in CI, so I expect that it will fail there as well.
> 
> After a bit of debugging, I think that the way `.bashrc` is sourced from `bashrc` is bogus: it does not correctly set the environment, when starting the container like above. Subsequently. when `run.sh` is executed, node is not in the path, and so the Theia app can't start. You can confirm by adding `echo "PATH: $PATH"` in `run.sh`
> 
> If instead one starts the container with command `bash`, then `.bashrc` is read and the environment is set correctly, and then one can start Theia.
> 
> Can you attempt to fix this?
> 
> BTW, I think this type of issue could be avoided if something else than `nvm` was used to pull node.js (e.g. install from archive and manually set PATH from image) ; then there would be no requirement for .bashrc to be read before being able to start Theia - $PATH as set in the image would be enough.
> 
> update: Theia starts if I comment-out `source $HOME/.bashrc` from `bashrc`.
> 
> update 2: I find suspicious, in `install_node.sh` line 18:
> 
> ```
> touch ~/.bash_profile 
> ```
> 
> I do not think it's necessary for `nvm`, and having this file will prevent `~/.profile` being sourced - that's the file that in turn sources `~/.bashrc`.

Earlier there was a problem when a terminal was started from within Theia. The installation of nvm also adds its setup to .bashrc.  As a result nvm setup was ran twice. I fixed that by checking if nvm environment vars were already set. If so I exited.  The exit in the nvm_setup script somehow caused the whole run.sh to be canceled. 

In the Dockerfile I use bash --rcfile /root/bashrc -i /root/run.sh. The "-i"  is very important otherwise indeed the env vars are not set correctly. Now that the "exit" has been removed from the "run.sh" it is again working I removed the changes that nvm automatically makes in the /root/.bashrc file. The Theia terminal is still working.

theia-rust-docker/README.md
dwjbosman: Updated readme

marcdumais-work:
@dwjbosman Thanks for the update - I'll test again.

marcdumais-work:
@dwjbosman I have confirmed that the container no longer immediately exits when I manually start it (without using the script) - thanks.

I noticed something afterwards: Looks like `$USER` environment variable is not set : 

```
root@358bcaff2148:~/workspace# cargo new hello-rust
error: Failed to create package `hello-rust` at `/root/workspace/hello-rust`

Caused by:
  could not determine the current user, please set $USER
```

I have not re-tested starting the container with the script, but I assume this is set somehow, in that case. Maybe add a default value in Dockerfile so that it's set as well when not using the script?

dwjbosman:
> @dwjbosman I have confirmed that the container no longer immediately exits when I manually start it (without using the script) - thanks.
> 
> I noticed something afterwards: Looks like `$USER` environment variable is not set :
> 
> ```
> root@358bcaff2148:~/workspace# cargo new hello-rust
> error: Failed to create package `hello-rust` at `/root/workspace/hello-rust`
> 
> Caused by:
>   could not determine the current user, please set $USER
> ```
> 
> I have not re-tested starting the container with the script, but I assume this is set somehow, in that case. Maybe add a default value in Dockerfile so that it's set as well when not using the script?

I have updated the Dockerfile to set user to "root" per default. This seems to work.

```
docker run --init --rm -d --name test_container -p 0.0.0.0:4000:3000 theia_rust_image
docker exec -ti test_container bash
source /root/rust/cargo/env
cargo new hello
```

I have "resquashed" the commits and added the signoff

dwjbosman:
So I see that there is another pull request adding rust support.  His approach is more similar to that of the other images. For me personally: for the theia-app image to be useful I need to be able to get to my local files easily. For that purpose the "ex" command was created. Furthermore the "ex" command allows you to run commands in the container from a terminal outside the container all while keeping the correct directories and environment.

The next.package.json of the other pull request seems interesting as it is using vscode rust debugger? I haven't tried it. Maybe we should merge the file from #209 ?

dwjbosman:
Once we finish this pull request I am planning to work on an image supporting rust for Embedded software development. This would add arm targets and openocd. Again the "ex" command would be used to pass the serial port to the container so that one can flash from inside the container.

marcdumais-work:
> I have "resquashed" the commits and added the signoff

Thanks @dwjbosman ; I'll try to find a bit of time to retest later today.

marcdumais-work:
> @marcdumais-work So I see that there is another pull request adding rust support. His approach is more similar to that of the other images. For me personally: for the theia-app image to be useful I need to be able to get to my local files easily. For that purpose the "ex" command was created. Furthermore the "ex" command allows you to run commands in the container from a terminal outside the container all while keeping the correct directories and environment.
> 

+1. And furthermore I think we can still use your version of this image the same way the other images in this repo are used, so I think we are well covered with your PR here. 

> The next.package.json of the other pull request seems interesting as it is using vscode rust debugger? I haven't tried it. Maybe we should merge the file from #209 ?

Merging a single file from a PR is not possible, AFAIK. 

Though I expect that the VS Code extension probably has some nice extra features and would be maintained better in the long term, I suggest we keep using the Theia extension in your PR for now, so we can merge ASAP. 

If the author of PR #206 or someone else cares to do so, I would welcome a follow-up PR that switches to the VS Code rust extension.

theia-rust-docker/next.package.json
```diff
@@ -0,0 +1,54 @@
...
+    "theia": {
+        "frontend": {
+            "config": {
+                "applicationName": "Theia Rust Example"
+            }
+        }
+    },
+    "dependencies": {
+        "@theia/core": "next",
+        "@theia/cpp-debug": "0.3.19",
```
marcdumais-work: We have made recent changes to `@theia/cpp-debug`, preparing to release it to NPM for the first time at the end of this month. The version has been stepped to `0.7.0', and so `0.3.19` is not found here, resulting in a failing build. I suggest you replace the specific version by '*' for now; once we release the first time, we'll also be able to release `next` versions for each commit, and so then we can just switch to `next` here as well. 

For now I have made the change locally so I can build and test.

theia-rust-docker/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+ENV CARGO_HOME /root/rust/cargo
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > /tmp/sh.rustup.rs && sh /tmp/sh.rustup.rs -y
+
+#install language server dependencies for theia
+RUN . /root/rust/cargo/env && rustup toolchain install $rust_channel
+RUN . /root/rust/cargo/env && rustup default $rust_channel
+RUN . /root/rust/cargo/env && rustup component add rust-src rust-docs rls rust-analysis rustfmt clippy
+
+RUN . /root/rust/cargo/env && cargo +$rust_channel install racer
```
marcdumais-work: This step fails for me now with a compilation error, using the default rust channel setup a few lines up (i.e. building "manually", without using the provided makefile). 

```
Step 26/34 : RUN . /root/rust/cargo/env && cargo +$rust_channel install racer
 ---> Running in e7ce53e2092f
    Updating crates.io index
 Downloading crates ...
  Downloaded racer v2.1.24
  Installing racer v2.1.24
 Downloading crates ...
  Downloaded derive_more v0.13.0
  Downloaded bitflags v1.1.0
  Downloaded clap v2.33.0
  Downloaded lazycell v1.2.1
  Downloaded log v0.4.8
  Downloaded env_logger v0.6.2
  Downloaded racer-cargo-metadata v0.1.1
  Downloaded humantime v1.2.0
  Downloaded rls-span v0.5.2
  Downloaded lazy_static v1.3.0
  Downloaded textwrap v0.11.0
  Downloaded atty v0.2.13
  Downloaded regex v1.2.0
  Downloaded strsim v0.8.0
  Downloaded cfg-if v0.1.9
  Downloaded racer-interner v0.1.0
  Downloaded quote v0.6.13
  Downloaded unicode-width v0.1.5
  Downloaded termcolor v1.0.5
  Downloaded vec_map v0.8.1
  Downloaded serde v1.0.98
  Downloaded proc-macro2 v0.4.30
  Downloaded ansi_term v0.11.0
  Downloaded syn v0.15.42
  Downloaded serde_json v1.0.40
  Downloaded quick-error v1.2.2
  Downloaded thread_local v0.3.6
  Downloaded serde_derive v1.0.98
  Downloaded rustc-ap-syntax v542.0.0
  Downloaded utf8-ranges v1.0.3
  Downloaded libc v0.2.60
  Downloaded aho-corasick v0.7.4
  Downloaded unicode-xid v0.1.0
  Downloaded memchr v2.2.1
  Downloaded itoa v0.4.4
  Downloaded regex-syntax v0.6.10
  Downloaded rustc-ap-serialize v542.0.0
  Downloaded ryu v1.0.0
  Downloaded rustc-ap-rustc_macros v542.0.0
  Downloaded rustc-ap-syntax_pos v542.0.0
  Downloaded rustc-ap-rustc_target v542.0.0
  Downloaded smallvec v0.6.10
  Downloaded rustc-ap-rustc_data_structures v542.0.0
  Downloaded rustc-ap-rustc_errors v542.0.0
  Downloaded scoped-tls v1.0.0
  Downloaded rustc-ap-rustc_lexer v542.0.0
  Downloaded ucd-util v0.1.5
  Downloaded itertools v0.8.0
  Downloaded synstructure v0.10.2
  Downloaded indexmap v1.0.2
  Downloaded jobserver v0.1.16
  Downloaded rustc-rayon v0.2.0
  Downloaded rustc-ap-arena v542.0.0
  Downloaded rustc-ap-graphviz v542.0.0
  Downloaded crossbeam-utils v0.6.6
  Downloaded annotate-snippets v0.6.1
  Downloaded rustc-rayon-core v0.2.0
  Downloaded ena v0.13.0
  Downloaded rustc-hash v1.0.1
  Downloaded stable_deref_trait v1.1.1
  Downloaded crossbeam-deque v0.2.0
  Downloaded either v1.5.2
  Downloaded parking_lot v0.7.1
  Downloaded num_cpus v1.10.1
  Downloaded byteorder v1.3.2
  Downloaded crossbeam-epoch v0.3.1
  Downloaded crossbeam-utils v0.2.2
  Downloaded lock_api v0.1.5
  Downloaded parking_lot_core v0.4.0
  Downloaded arrayvec v0.4.11
  Downloaded memoffset v0.2.1
  Downloaded nodrop v0.1.13
  Downloaded scopeguard v0.3.3
  Downloaded owning_ref v0.4.0
  Downloaded rand v0.6.5
  Downloaded rustc_version v0.2.3
  Downloaded rand_chacha v0.1.1
  Downloaded rand_core v0.4.0
  Downloaded rand_xorshift v0.1.1
  Downloaded rand_os v0.1.3
  Downloaded rand_isaac v0.1.1
  Downloaded autocfg v0.1.5
  Downloaded rand_pcg v0.1.2
  Downloaded rand_hc v0.1.0
  Downloaded rand_core v0.3.1
  Downloaded rand_jitter v0.1.4
  Downloaded semver v0.9.0
  Downloaded semver-parser v0.7.0
   Compiling autocfg v0.1.5
   Compiling arrayvec v0.4.11
   Compiling semver-parser v0.7.0
   Compiling libc v0.2.60
   Compiling cfg-if v0.1.9
   Compiling rand_core v0.4.0
   Compiling nodrop v0.1.13
   Compiling lazy_static v1.3.0
   Compiling memoffset v0.2.1
   Compiling proc-macro2 v0.4.30
   Compiling scopeguard v0.3.3
   Compiling rustc-rayon-core v0.2.0
   Compiling log v0.4.8
   Compiling byteorder v1.3.2
   Compiling unicode-xid v0.1.0
   Compiling stable_deref_trait v1.1.1
   Compiling smallvec v0.6.10
   Compiling syn v0.15.42
   Compiling rustc-rayon v0.2.0
   Compiling indexmap v1.0.2
   Compiling either v1.5.2
   Compiling rustc-ap-graphviz v542.0.0
   Compiling memchr v2.2.1
   Compiling serde v1.0.98
   Compiling bitflags v1.1.0
   Compiling ryu v1.0.0
   Compiling ucd-util v0.1.5
   Compiling unicode-width v0.1.5
   Compiling scoped-tls v1.0.0
   Compiling rustc-ap-rustc_target v542.0.0
   Compiling annotate-snippets v0.6.1
   Compiling rustc-ap-syntax v542.0.0
   Compiling quick-error v1.2.2
   Compiling itoa v0.4.4
   Compiling termcolor v1.0.5
   Compiling utf8-ranges v1.0.3
   Compiling rustc-ap-rustc_lexer v542.0.0
   Compiling strsim v0.8.0
   Compiling vec_map v0.8.1
   Compiling ansi_term v0.11.0
error[E0658]: use of unstable library feature 'rustc_private': mainly needed for compiler internals
   --> /root/rust/cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-rustc_lexer-542.0.0/src/lib.rs:689:33
    |
689 |             || (c > '\x7f' && c.is_xid_start())
    |                                 ^^^^^^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/27812
    = help: add `#![feature(rustc_private)]` to the crate attributes to enable

error[E0658]: use of unstable library feature 'rustc_private': mainly needed for compiler internals
   --> /root/rust/cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-rustc_lexer-542.0.0/src/lib.rs:708:33
    |
708 |             || (c > '\x7f' && c.is_xid_continue())
    |                                 ^^^^^^^^^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/27812
    = help: add `#![feature(rustc_private)]` to the crate attributes to enable

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0658`.
error: Could not compile `rustc-ap-rustc_lexer`.
warning: build failed, waiting for other jobs to finish...
error: failed to compile `racer v2.1.24`, intermediate artifacts can be found at `/tmp/cargo-installcDtghX`

Caused by:
  build failed
The command '/bin/sh -c . /root/rust/cargo/env && cargo +$rust_channel install racer' returned a non-zero code: 101
```

tests/package.json
```diff
@@ -4,14 +4,15 @@
   "version": "0.0.1",
   "scripts": {
     "test": "wdio wdio.conf.js --theia-port 4000",
-    "test-app-theia": "wdio wdio.theia-docker.conf.js --theia-port 4000",
+    "test-app-theia": "wdio wdio.theia-docker.conf.js --theia-port 3000",
```
marcdumais-work: Please change-back port to 4000

theia-rust-docker/Dockerfile
marcdumais-work: update: I can't build as well with the `stable` channel and also if I use the makefile. Maybe something was broken Dockerfile?

tests/package.json
dwjbosman: ok

theia-rust-docker/next.package.json
dwjbosman: ok

theia-rust-docker/Dockerfile
dwjbosman: It seems that there was an update in rust somewhere. The Racer component no longer compiles.  Strange thing is that this happens even while setting the rust channel to a specific date. Have to look into this.

marcdumais-work:
> Once we finish this pull request I am planning to work on an image supporting rust for Embedded software development. This would add arm targets and openocd. Again the "ex" command would be used to pass the serial port to the container so that one can flash from inside the container.

Sounds interesting, though I do not know much about it.  We have a few people from ARM involved in Theia, BTW. If technically possible, it would be good to share some of the infrastructure between what's contributed in this PR here and the new one.

theia-rust-docker/Dockerfile
marcdumais-work: I just tried again, and now the image builds. Strange.

marcdumais-work: Ah, maybe the build working again has to do with the commit you added.

marcdumais-work:
@dwjbosman I will be going on vacation at the end of this week. If you have time to squash the commits one last time, in the next couple of days, I'll test quickly and hopefully be able merge this PR.

theia-rust-docker/Dockerfile
dwjbosman: I temporarily disabled installing Racer.

dwjbosman:
> @dwjbosman I will be going on vacation at the end of this week. If you have time to squash the commits one last time, in the next couple of days, I'll test quickly and hopefully be able merge this PR.

I'll try to do it this tonight.

dwjbosman:
> > @dwjbosman I will be going on vacation at the end of this week. If you have time to squash the commits one last time, in the next couple of days, I'll test quickly and hopefully be able merge this PR.
> 
> I'll try to do it this tonight.

Fixed Racer compilation issue.
Squashed the commits.

theia-rust-docker/example_workspace/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[workspace]
+
+members = [
+    "adder",
+]
```
vince-fugnitto: [minor] please include a new line.

theia-rust-docker/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+<pre>ex_bash "cargo --list"</pre>
+
+## Example project
+
+A very simple rust example project is included in the example_project folder. It contains a build task and debug configuration which uses rust-gdb.
+
+## How does it work?
+
+The ex\* scripts check if a container exists and if so reuses it. If not, the container is created. A command is ran inside the container to create a user corresponding to the user calling the ex script. If ex_theia is ran, the bashrc script is sourced after which the run.sh script is ran inside the container. The bashrc script changes the dir to the same location from which the ex script was ran (via the CONTAINER_START_PATH environment variable). This is done only if the current path is within the users mounted HOME dir. Next the bashrc script sources nvm_setup.sh which activates a specific node version and adds yarn to the path. Lastly it sets up the rust build environment through RUSTUP_HOME and CARGO_HOME. Once the environent is sourced the run.sh script can actually start Theia.
```
vince-fugnitto: [minor] typo `environent` -> `environment`.

commit_screenshots.sh
```diff
@@ -0,0 +1,48 @@
...
+git config user.name "Travis CI"
+git config user.email "$COMMIT_AUTHOR_EMAIL"
+
+git fetch --depth 1 origin $TARGET_BRANCH:$TARGET_BRANCH
+git checkout $TARGET_BRANCH
+
+#empty the branch, ignore error if alrady empty
+git rm -r . || true
+
+#add all png files
```
vince-fugnitto: [minor] there are a few comments throughout the PR where we include spaces at the beginning, and others where we do not, can we be consistent throughout and include a space? (ex: `# add all...` instead of `#add all`).

theia-rust-docker/next.package.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "dependencies": {
+        "@theia/core": "next",
+        "@theia/cpp-debug": "*",
+        "@theia/editor": "next",
+        "@theia/editorconfig": "next",
+        "@theia/editor-preview": "next",
+        "@theia/file-search": "next",
+        "@theia/filesystem": "next",
+        "@theia/git": "next",
```
vince-fugnitto: [minor] there is some leftover whitespace here. (editing within VSCode)

<div align='left'>

Screen Shot 2019-08-12 at 10 35 52 AM


</div>

dwjbosman: ok

commit_screenshots.sh
dwjbosman: ok

theia-rust-docker/README.md
dwjbosman: ok

vince-fugnitto:
Please ping me when it's ready for re-review :)

theia-rust-docker/example_workspace/Cargo.toml
dwjbosman: ok

dwjbosman:
@vince-fugnitto Done resolving your comments

vince-fugnitto:
Thank you very much for your contribution, really nice work!
I hope we can see many more contributions from you in the future :)

PR #205, opened 2019-07-19
[full] add rust toolchain to theia-full-docker

satishbabariya:


marcdumais-work:
Hi @satishbabariya ,

No worries about the CI failing - it's the case most of the time for PR originating from forks of this repo. I'll validate that everything builds correctly before merging.

marcdumais-work:
@satishbabariya Can you please squash the two commits into a single one, and add "[full] " in front of the commit message, so it's clear the change applies to the "full" image.

marcdumais-work:
@satishbabariya BTW, you might be interested in a recent contribution to this repo, still under review, that adds a dedicated rust docker image: https://github.com/theia-ide/theia-apps/pull/200

It's quite nice, and even supports debugging.

satishbabariya:
@marcdumais-work, done and I'll look into #200

satishbabariya:
@marcdumais-work I've checked out #200 and it still lacking and can be much simplified further.
I'd dropped pr on it. for another clean implementation of theia rust docker with vscode extension for next and theia-rust-extension for the letest image.

marcdumais-work:
@satishbabariya Thanks for squashing the commits. You have duplicated "sign-offs" ATM as part of your commit message - could you remove one first one please? Otherwise I think this PR is ready to merge.

satishbabariya:
@marcdumais-work done.

marcdumais-work:
@satishbabariya one more thing: I think there may be a permission issue, or alternatively I am doing something wrong. I was trying to minimally test the rust toolchain you added using this: https://www.rust-lang.org/learn/get-started

I am able to create a rust project using command `cargo new hello-rust`, but when I try to execute it with `cargo run`, I get the following error: 

```bash
theia@c447d1baf050:/home/project/hello-rust$ cargo run
error: failed to acquire package cache lock

Caused by:
  failed to open: /usr/local/cargo/.package-cache

Caused by:
  Permission denied (os error 13)
```

It looks like `cargo run` expects to be able to write in path ` /usr/local/cargo/.package-cache`, which is owned by root. Maybe there is an environment variable that can be set to tell it to use a folder in the `theia` user's home instead?

theia-full-docker/Dockerfile
```diff
@@ -139,6 +139,17 @@ RUN apt-get -y install ruby ruby-dev zlib1g-dev && \
     gem install solargraph
 
 
+# Rust
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH \
+    USER=$USER
```
marcdumais-work: Another small thing I noticed: it did not seem to cause an issue, but in the container, environment variable `$USER` is set, but to nothing. I think it's because $USER is not set in Dockerfile. 

If I unset the variable, then `cargo new hello-rust` fails. So I guess it needs to be set, even if the value is empty. Was this intended on your part or did you want it to have a specific value?

satishbabariya: done, also updated #209, also final thought I think `go get also will not work in this image` it's the same case with permission, I'll drop pr on it also so np.

satishbabariya:
@marcdumais-work, updated dockerfile, redy to test now, also updated #209, also final thought I think `go get` also will not work in this image it's the same case with permission, I'll drop pr on it also so np.

marcdumais-work:
> updated dockerfile, redy to test now

Yes, it works now.

marcdumais-work:
Also now $USER is defined in the container, with value "root".

PR #209, opened 2019-07-24
[rust] add theia-rust-docker

satishbabariya:


marcdumais-work:
cc: @dwjbosman (author of PR #200)

Thanks @satishbabariya for this PR. 

Your version of the rust image is similar to the other docker images we have so far in this repo for specific languages: a simple, straightforward docker image, in which one can try a Theia-based IDE to do some rust development or demos. 

The other rust image, under review in PR #200, is a more elaborate solution, that can to be used as a dockerized desktop development environment for rust, adjusting the permissions so they align both in the container and on the host, and as well provides scripts to help the user build and run the docker image. When the image is started with specific parameters, it permits debugging rust program as well. I think the image can be used as the Cloud, but then the closer integration to the user's host would be lost. There is added complexity, and a user would need to carefully follow instructions from the image's README to get all the benefits.

So I think that both PRs have merit and either one could be seen as preferable depending on the specific use-case. However I would prefer if we only had at most one image per language (a "latest" and a "next" versions for each is ok), since CI already takes quite long; the Travis CI executors we are organization-wide, and so long CI here may increasingly delay CI for the main Theia repo. 

You [mentioned](https://github.com/theia-ide/theia-apps/pull/205#issuecomment-513563950) elsewhere that you found PR #200 "lacking", but did not elaborate on the specifics, though I understood the added "bells and whistles", that you may not need, could be part of it. Do you think we could instead work on improving PR #200 so that it's fit for your use-cases as well? Or alternatively make improvements in a second pass once it's merged?

satishbabariya:
@marcdumais-work we can use this image as the base image and further update if required.

marcdumais-work:
> @marcdumais-work we can use this image as the base image and further update if required.

@satishbabariya  Just so I am sure I understand correctly what you mean: by "this image" do you mean the one in this PR here or the one from PR #200?

satishbabariya:
> > @marcdumais-work we can use this image as the base image and further update if required.
> 
> @satishbabariya Just so I am sure I understand correctly what you mean: by "this image" do you mean the one in this PR here or the one from PR #200?

@marcdumais-work I'm referencing this pr #209

vince-fugnitto:
Closing in favor of #200 for rust support.
Thank you for your contribution!

PR #221, opened 2019-08-26
Permissions

dwjbosman:
There is a problem with the Rust app. Currently the files in the cargo folder are owned by "root". Thus the user is not able to install crates. Even more the cache folder is also not writeable. This causes errors when the RLS wants to cache. 

The solution was to add a "rust" group and make all the files in the rust folder (cargo and rustup) part of this group and group writable.  When the user creates the container the specific user is created inside the container and added to the rust group. The users umask is set to 0002 so that any changes made to the cargo or cache folders are also group writeable. (Drawback) Note that this will also make all other files created by the user (eg. source code) group writeable.

I still have to run some tests


dwjbosman:
> @dwjbosman do you mind squashing your commits into a single commit and
> including the `sign-off` (reason for the DCO failed check)?

ok, done.

Issue #214, opened 2019-07-25
[CI] only build/test images that have been updated by the commit


vince-fugnitto:
**Description**

Only build and test image(s) which are directly affected by the commit.

**Problem**

For each pull-request `travis` builds and tests every single image in the repository irregardless of if the image has actually been updated which is both **time and resource consuming**.

**Solution**

The proposal is to include [**`conditional builds/stages`**](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/) to determine which image(s) should be built and tested.

The idea is to do so by either of the following [**`methods`**](https://docs.travis-ci.com/user/conditions-v1):
1. add a `tag` which can be set through a post-commit hook
2. make use of the `commit-message` which can include which images are updated and be a standard part of the review process.

marcdumais-work:
(test - please ignore)

@theia-ide/ip-help testing the new team in a mention

dwjbosman:
Wouldn't it be possible to:
1. Use git to determine which files were changed
2. Based on the changed files determine the application folders and theia apps that have changed.
3. Then, based on the changed theia apps, run the build/test scripts.

That way a user wouldn't have to do anything special. 

A bash script should be able to do step 1 and 2. Call the script from Travis.

dwjbosman:
Hi, I tried to address this using the TRAVIS_EVENT_TYPE environment variable. Please see the pull request comments for more info.

marcdumais-work:
Hi @dwjbosman ,

I think it would be simpler for both us and you, if we were to give you write access to this repo here; this way your PRs can originate from the main repo and be covered by the GitHub token, reducing build failures. WDYT?

dwjbosman:
That's fine. Shall I then recreate the pull requests?

marcdumais-work:
@dwjbosman I have added you as a collaborator in this repo - you should receive an email notification with a link to accept

> Shall I then recreate the pull requests?

Yes, I think that's the way to go - thanks.

dwjbosman:
I have not received an activation email (checked the spam etc.). Can you resend it?

marcdumais-work:
> I have not received an activation email (checked the spam etc.). Can you resend it?

The invitation is no longer pending - you are listed as member with write access. Please try pushing your "build_on_changes" branch to the main repo and let me know if it doesn't work. If it does work, everything is good.

dwjbosman:
After merging the build immediately failed :)

There was an error:
[...]

PR #222, opened 2019-08-26
Build on changes

dwjbosman:
A proposed solution to issue #214. On each commit only those images will be built/tested that were changed.


dwjbosman:
> @dwjbosman do you mind squashing your commits into a single commit and
> including the `sign-off` (reason for the DCO failed check)?

ok

dwjbosman:
I used TRAVIS_EVENT_TYPE to determine if the build was started through a commit or via the dashboard (more options -> trigger build)

https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

Issue #223, opened 2019-08-27
[full, cpp] clangd, clang-tidy packages not found


marcdumais-work:
Our `cpp` images are not building ATM. Looks like we need to update the clangd/clang-tidy Debian packages we pull:

```
E: Unable to locate package clang-tools-9
E: Unable to locate package clang-tidy-9
```

It might be because LLVM have switched to version 10 as the bleeding edge dev version and we need to use that one. @lmcbout mentioned that now `clangd` has it's own Debian package instead of being part of `clang-tools`, so we may need to install that new package as well. 

See build errors: 
https://travis-ci.org/theia-ide/theia-apps/jobs/577117058#L4013
https://travis-ci.org/theia-ide/theia-apps/jobs/577117072#L1433

vince-fugnitto:
Closed in #232

PR #225, opened 2019-08-28
[full] [cpp] Increment clang from version 9 to version 10 for LLVM PP…

daball:
…A. Resolves issue #223.

Installation proceeds past Step 18/39, with console output:

`The following NEW packages will be installed:
  binfmt-support binutils binutils-common binutils-x86-64-linux-gnu clang-10
  clang-tools-10 file gcc-7-base lib32gcc1 lib32stdc++6 libasan4 libatomic1
  libbinutils libc-dev-bin libc6-dev libc6-i386 libcilkrts5
  libclang-common-10-dev libclang-cpp1-10 libclang1-10 libffi-dev libgc1c2
  libgcc-7-dev libgomp1 libitm1 libllvm10 liblsan0 libmagic-mgc libmagic1
  libmpdec2 libmpx2 libobjc-7-dev libobjc4 libomp-10-dev libomp5-10 libpfm4
  libpipeline1 libpython3-stdlib libpython3.6-minimal libpython3.6-stdlib
  libquadmath0 libstdc++-7-dev libtinfo-dev libtsan0 libubsan0 linux-libc-dev
  llvm-10 llvm-10-dev llvm-10-runtime manpages manpages-dev mime-support
  python3 python3-minimal python3.6 python3.6-minimal`

### DIFF
--- a/theia-cpp-docker/Dockerfile
+++ b/theia-cpp-docker/Dockerfile
@@ -97,11 +97,11 @@ RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
     echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main" > /etc/apt/sources.list.d/llvm.list && \
     apt-get update && \
     apt-get install -y cmake \
-                       clang-tools-9 \
-                       clang-tidy-9 \
+                       clang-tools-10 \
+                       clang-tidy-10 \
                        gdb && \
-    ln -s /usr/bin/clangd-9 /usr/bin/clangd && \
-    ln -s /usr/bin/clang-tidy-9 /usr/bin/clang-tidy && \
+    ln -s /usr/bin/clangd-10 /usr/bin/clangd && \
+    ln -s /usr/bin/clang-tidy-10 /usr/bin/clang-tidy && \
     rm -rf /var/lib/apt/lists/* 
 
 ## User account
--- a/theia-full-docker/Dockerfile
+++ b/theia-full-docker/Dockerfile
@@ -118,8 +118,8 @@ RUN apt-get update && apt-get -y install openjdk-8-jdk maven gradle
 # public LLVM PPA, development version of LLVM
 RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
     echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main" > /etc/apt/sources.list.d/llvm.list && \
-    apt-get update && apt-get install -y clang-tools-9 && \
-    ln -s /usr/bin/clangd-9 /usr/bin/clangd
+    apt-get update && apt-get install -y clang-tools-10 && \
+    ln -s /usr/bin/clangd-10 /usr/bin/clangd
 
 #Python 2
 RUN apt-get update && apt-get install -y python python-pip && \


marcdumais-work:
> There did seem to be an intermittent issue in the CI build; possibly an API rate limit exceeded. Per https://developer.github.com/v3/#abuse-rate-limits, CI log has errors like:
> 

Indeed - we have a GH Token setup in the Travis' account config to avoid that, but for security reason Travis does not use it for PRs that originate from a fork, so it only helps committers to this repo. 

So it's nothing for contributors to worry-about - we build locally before merging to insure all is good.

> These were the same changes I applied. It did build locally after the changes were applied.

Correct, it builds. But as far as I understand, starting with version 10 `clangd` has its own Debian package and is no longer included in the `clang-tools` package, so now it needs to be installed as well, else no Language Server will be present in the image. 

@lmcbout has taken-over; he will start from your PR's commit and add what's missing, including a thing or two not directly related to the issue you were fixing. We'll preserve your original commit, including authorship.

Thanks again for this contribution, and feel free to look at the current issues if you'd like to contribute some more.

daball:
> Indeed - we have a GH Token setup in the Travis' account config to avoid that, but for security reason Travis does not use it for PRs that originate from a fork, so it only helps committers to this repo.

Nice.

> So it's nothing for contributors to worry-about - we build locally before merging to insure all is good.

Same. I didn't have any past context or project direction to determine if the commit was 100% on point. I merely needed to get it to build. I wasn't actually aware of how active the project was.

> Correct, it builds. But as far as I understand, starting with version 10 `clangd` has its own Debian package and is no longer included in the `clang-tools` package, so now it needs to be installed as well, else no Language Server will be present in the image.

I'll admit, this certainly fits into the million things I don't like about C/C++ and friends box. I was merely trying to build the `theia-full` image. I wasn't able to run any sort of acceptance tests or to check the features out at all.

> @lmcbout has taken-over; he will start from your PR's commit and add what's missing, including a thing or two not directly related to the issue you were fixing. We'll preserve your original commit, including authorship.

Thanks! ;) Somebody that's more familiar with C++ should handle the rest.

> Thanks again for this contribution, and feel free to look at the current issues if you'd like to contribute some more.

You're welcome.

I like where the project is going. I love VS Code already, and this thing looks just like it. In fact, I used VS Code when I made these small changes to the code.

I haven't actually used it yet though. I installed the Docker image and that's about as far as I got. I kind of stopped as I was more or less trying to figure out if I need to be able to spin up and down containers and to maybe have a secondary UI for handling provisioning in a more approachable way. Particularly, I aim to use a private GitLab server for hosting, and it would be nice to start with a view that contains each of my private/public/group repos and any connected Theia instances; either to join a peer instance or to spin up a separate instance. I don't actually know that it supports peer programming straight out, but this would be a nice feature. The glue in the middleware should spin up an instance, possibly using Kubernetes, clone the repo, and then dive into the Docker-configured IDE. I would expect the IDE should be able to commit, push, and pull changes to the local and remote, and examine GitLab/Jira/etc. issues at some point. And then all of it behind the same SSO/LDAP/authorization wall.

Aside from the secondary UI, I'd be interested in working on an extension I've been kicking around for a while. I would like to develop a code generation tool. Mostly because I'm tired of monolithic generation tools which generate lots of files and folders, and I'd rather generate one thing at a time. I've never created a tool extension though, so there's lots to keep me busy.

Perhaps I could help work on some issues as they come up, when I have some extra time. Thanks for being awesome!

PR #232, opened 2019-08-29
Clangd10 take2

marcdumais-work:
Re-trying to merge the content of PR #229 in a way that preserves the original contained commits.


marcdumais-work:
CI is passing for theia-full and theia-cpp, so looks good.

PR #236, opened 2019-09-06
Travis should only build/test images that have changes

dwjbosman:
A proposed solution to issue #214. On each commit only those images will be built/tested that were changed.

I used TRAVIS_EVENT_TYPE to determine if the build was started through a commit or via the dashboard (more options -> trigger build)

https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

    If a commit is done only the specific image is built/tested.
    If "trigger build" is clicked via the travis dashboard, all images are built/tested

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,18 +32,15 @@ env:
   - NPM_TAG=next IMAGE_NAME=theia-dart NODE_VERSION=10
 
 install:
-  - cd "$IMAGE_NAME-docker"
-  - IMAGE="theiaide/$IMAGE_NAME"
-  - IMAGE_TAG="$IMAGE":$(npm view "@theia/core@$NPM_TAG" version)
-  - echo $IMAGE_TAG
   # this makes it possible to see output of the docker build as it unfolds:
   # from https://stackoverflow.com/questions/43918874/how-to-increase-no-activity-wait-time-in-travis-ci
   - "travis_wait 30 sleep 1800 &"
-  - docker build --build-arg "version=$NPM_TAG" --build-arg "NODE_VERSION=$NODE_VERSION" --build-arg "GITHUB_TOKEN=$GITHUB_TOKEN" . -t "$IMAGE_TAG" --no-cache
-  - docker tag "$IMAGE_TAG" "$IMAGE:$NPM_TAG"
-  - docker run --init -d -p 0.0.0.0:4000:3000 "$IMAGE_TAG"
+  # check if there were changes since the previous commit in this image
+  # if not (script returns 137) then the build process is terminated
+  - ./check_changed.sh $IMAGE_NAME ; RETURN_CODE=$? ; if [ $RETURN_CODE -eq 137 ]; then travis_terminate 0; elif [ $RETURN_CODE -ne 0 ]; then travis_terminate $RETURN_CODE; fi
+  - ./build_container.sh $NPM_TAG $IMAGE_NAME $NODE_VERSION
 
-script: cd ../tests; yarn; yarn test-app-$IMAGE_NAME
+script: cd tests; yarn; yarn test-app-$IMAGE_NAME
 
 after_failure:
   script: cd $TRAVIS_BUILD_DIR && ./commit_screenshots.sh
--- a/build_container.sh
+++ b/build_container.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+
+# this script is called by Travis to build the Docker image
+NPM_TAG=$1
+IMAGE_NAME=$2
+NODEVERSION=$3
+
+cd "$IMAGE_NAME-docker"
+
+IMAGE="theiaide/$IMAGE_NAME"
+IMAGE_TAG="$IMAGE":$(npm view "@theia/core@$NPM_TAG" version)
+echo $IMAGE_TAG
+docker build --build-arg "version=$NPM_TAG" --build-arg "NODE_VERSION=$NODE_VERSION" --build-arg "GITHUB_TOKEN=$GITHUB_TOKEN" . -t "$IMAGE_TAG" --no-cache
+docker tag "$IMAGE_TAG" "$IMAGE:$NPM_TAG"
+docker run --init -d -p 0.0.0.0:4000:3000 "$IMAGE_TAG"
--- a/check_changed.sh
+++ b/check_changed.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+set -e
+# This script is called by Travis during the install step.
+# It returns 1 if no files where changed. In that case
+# no further building/test is required for this image
+
+IMAGE_NAME=$1
+
+echo "Travis event type: $TRAVIS_EVENT_TYPE"
+
+if [ "$TRAVIS_EVENT_TYPE" == "api" ]
+then
+    # trigger via travis dashboard
+    # test all
+    echo "Trigger all tests"
+    exit 0
+fi
+
+cd "$IMAGE_NAME-docker"
+
+CHANGED_FILES=$(git diff --name-status HEAD~1...HEAD .)
+if [ -z "$CHANGED_FILES" ]
+then
+    # nothing changed, skip building
+    echo "No changes in $IMAGE_NAME, terminate"
+   
+    # this indicates to the parent script that the build can be terminated
+    exit 137
+fi
+
+echo "There were changes in $IMAGE_NAME changes: $CHANGED_FILES"
+exit 0


dwjbosman:
This pull request has been recreated. It is now part of the original repo.

build_container.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -e
+
+# this script is called by Travis to build the Docker image
+NPM_TAG=$1
+IMAGE_NAME=$2
+NODEVERSION=$3
+
+cd "$IMAGE_NAME-docker"
+
```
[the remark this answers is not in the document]
dwjbosman: I corrected this by using an absolute path in the script section.

check_changed.sh
```diff
@@ -0,0 +1,34 @@
...
+
+cd "$IMAGE_NAME-docker"
+
+CHANGED_FILES=$(git diff --name-status $TRAVIS_BRANCH~1...$TRAVIS_BRANCH .)
+if [ -z "$CHANGED_FILES" ]
+then
+    # nothing changed, skip building
+    exit 1 
+fi
+echo "There were changes in $IMAGE_NAME changes: $CHANGED_FILES"
```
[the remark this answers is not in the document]
dwjbosman: It was a bit messy because the check_changes script used a return code to indicate if something changed or not. If a real error occurred in the script (eg. git timeout) this was mistakenly interpreted as "no changes".

I moved 'travis_terminate'  into the script. No more mishandling of bash return codes.

.travis.yml
```diff
@@ -2,7 +2,7 @@ services:
   - docker
 
 git:
-  depth: 1
+  depth: 2
```
vince-fugnitto: Can you explain a little why a `git.depth` of 2 is used?
Is it sufficient in all cases? (ex: if a PR contains multiple commits when appropriate).

theia-cpp-docker/Dockerfile
```diff
@@ -97,11 +97,12 @@ RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
...
     apt-get install -y cmake \
-                       clang-tools-9 \
-                       clang-tidy-9 \
+                       clang-tools-10 \
+                       clangd-10 \
+                       clang-tidy-10 \
                        gdb && \
-    ln -s /usr/bin/clangd-9 /usr/bin/clangd && \
-    ln -s /usr/bin/clang-tidy-9 /usr/bin/clang-tidy && \
+    ln -s /usr/bin/clangd-10 /usr/bin/clangd && \
```
vince-fugnitto: I don't believe this is part of your intended changes, the change is present in #232

theia-full-docker/Dockerfile
```diff
@@ -118,8 +118,9 @@ RUN apt-get update && apt-get -y install openjdk-8-jdk maven gradle
 # public LLVM PPA, development version of LLVM
 RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
     echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main" > /etc/apt/sources.list.d/llvm.list && \
-    apt-get update && apt-get install -y clang-tools-9 && \
-    ln -s /usr/bin/clangd-9 /usr/bin/clangd
+    apt-get update && apt-get install -y clang-tools-10 clangd-10 clang-tidy-10 && \
```
vince-fugnitto: Same as previous comment, change is part of #232

theia-cpp-docker/Dockerfile
dwjbosman: Ok

theia-full-docker/Dockerfile
dwjbosman: Ok

.travis.yml
dwjbosman: A depth of 2 ensures that the last two commits to a specific branch are available. 
The check_changes script then retrieves the changes between these 2 commits and filters. Only changes for a certain directory tree are returned:

`git diff --name-status $TRAVIS_BRANCH~1...$TRAVIS_BRANCH .`

So every time a new commit is pushed to a pull request and it contains changes to files inside the "$IMAGE_NAME-docker" folder then the image will be rebuilt.

dwjbosman:
> @dwjbosman do you mind squashing your commits when ready for another round of review?
> 
> I have a few high-level questions (which will likely need to be tested):
> 
>     * will it support multiple-commit PRs?

Yes I think so. It will check for differences since the last commit in a specific branch.

>     * if a PR consists of changes outside of any image (ex: `theia-apps/README.md`), I assume no image is built. However, should we fallback to testing at least one image? (ex: `theia-full-docker`)

Correct. 

I could add something that checks: If all changes files reside in non "*-docker" folders then build theia-full-docker,

>     * I assume skipped or unbuilt images will not result in a failed CI check

Correct, in .travis.yml:

`travis_terminate 0` ensures that the build stops but is marked as succeeded
.

lmcbout:
I did a test. I created 3 commits on top of your PR, push it and observe the result.
First commit, modify file in "`cpp app`", second commit in "`java app`" and third in "`go app`".
PR: https://github.com/theia-ide/theia-apps/pull/240
Travis build: https://travis-ci.org/theia-ide/theia-apps/builds/584273509

Observations:
1- The script only considered the file modified in the first commit  (`cpp app`) which affected one app only.
  The files modified in the 2 others commits (`java and go`) were not detected
2- Travis tag the end-result only on the last commit , but did not consider the files in that commit (go app)

marcdumais-work:
> Yes I think so. It will check for differences since the last commit in a specific branch.

I think I see what you mean; if one gradually pushes new commit on a PR branch, the modified images would be tested for each commit, over time? 

But I think there is an assumption that CI tests the whole PR however. and nothing prevents a contributor working locally or on a fork, and then submit a PR containing multiple commits at once. I think there may be a simple way - let me check something.

marcdumais-work:
@dwjbosman  I think it's possible to slightly simplify the `check_changed.sh` script and have the differences for all the commits in a PR, not just the top one. Turns out that GitHub/Travis already does most of that work; when Travis does its thing, the first commit on the test branch is a merge commit that contains all changes for the PR. Not sure of all the details but I think it may even merge the head of master into the PR branch, insuring valid tests in the case the PR is not based on latest master.

So I think it's not necessary to compare branches, just something like `git diff --name-only HEAD HEAD~1 ` should do the trick.

See my test PR #241 where I add 3 files to the root or the repo, `file1`, `file2` and `file3` each in a different commit. Here's the result of the `git diff` command: https://travis-ci.org/theia-ide/theia-apps/jobs/584695052#L456-L460 , showing all these changes.

dwjbosman:
> So I think it's not necessary to compare branches, just something like `git diff --name-only HEAD HEAD~1 ` should do the trick.

Thank you for your analysis. I was just debugging and found out that Travis is already doing a merge. I updated the script. PR #241 seems to correctly build only the changed images now.  See build 1238 ( https://travis-ci.org/theia-ide/theia-apps/builds/584819505)

marcdumais-work:
One thing that I think would have been nice for this PR, would be to have, on the Travis result page,  a different status for the images that were skipped, in a given CI run, because the script thinks they did not need building/testing. 

Unfortunately it seems that Travis does not support that: the only status available are success and failure, as far as I can tell. ATM the images that are skipped are reported as having succeeded, which is the best of the 2 choices I think.

dwjbosman:
It seems to be possible to give each job a name. If it is somehow possible to name all jobs of non-tested   images to "skipped" we could use that.

marcdumais-work:
> It seems to be possible to give each job a name. If it is somehow possible to name all jobs of non-tested images to "skipped" we could use that.

Sounds interesting; if you have time to give this a try, I'd like to see if this helps. Else I think looking at build time will give us a good clue which images been rebuilt/tested and which were skipped.

.travis.yml
lmcbout: No need to change the depth anymore , a depth of 1 is fine since we do a diff between HEAD and HEAD~1

dwjbosman: Ok, will update it

dwjbosman:
> It seems to be possible to give each job a name. If it is somehow possible to name all jobs of non-tested images to "skipped" we could use that.

Unfortunately the job names cannot be specified "dynamically" during the build. So we cannot use this. As suggested, the best approach is to look at the build times.

theia-cpp-docker/Dockerfile
dwjbosman: Hmm, I did a quick test. --depth 2 is required:

```
git clone --depth 1 git@github.com:theia-ide/theia-apps.git
cd theia-apps
$ git diff HEAD HEAD~1
fatal: ambiguous argument 'HEAD~1': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

With "-depth 2" this command works as expected. A depth of one ensures that the HEAD is available, a depth of 2 also ensures that we can compare with the previous commit.

.travis.yml
dwjbosman: Hmm, I did a quick test. --depth 2 is required:

```
git clone --depth 1 git@github.com:theia-ide/theia-apps.git
cd theia-apps
$ git diff HEAD HEAD~1
fatal: ambiguous argument 'HEAD~1': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

With "-depth 2" this command works as expected. A depth of one ensures that the HEAD is available, a depth of 2 also ensures that we can compare with the previous commit.

lmcbout: I tried your code plus 3 commits (different apps) on top of it ( https://travis-ci.org/theia-ide/theia-apps/builds/586164681 ) and I put depth = 1.
Everything seems fine. Can you give me an example when I need depth=2
Thanks

marcdumais-work: I think `depth=1` is enough - the way Travis works you'll still have all you need. See this example: https://travis-ci.org/theia-ide/theia-apps/jobs/584695052#L444-L455

That being said, the extra resources used with `depth=2` are negligible.

dwjbosman: Ok, reverted

Issue #246, opened 2019-09-27
[travis] the 'theia-full-docker' image should be tested when no images are tested


vince-fugnitto:
**Description**

With the introduction of PR #236, only images which have changes are built and tested during CI. I propose that if a PR introduces changes that do not directly affect any image, that the base `theia-full-docker` image is at least tested.

At the moment, this is not the case: #245 - [travis results](https://travis-ci.org/theia-ide/theia-apps/builds/590554565?utm_source=github_status&utm_medium=notification)

@dwjbosman is this something you'd like to handle?

dwjbosman:
Ok

vince-fugnitto:
Closed by #247

PR #247, opened 2019-09-28
Build theia-docker in case no specific images were changed

dwjbosman:
Solution for #246

### DIFF
--- a/check_changed.sh
+++ b/check_changed.sh
@@ -16,9 +16,39 @@ then
     exit 0
 fi
 
-cd "$IMAGE_NAME-docker"
 
-CHANGED_FILES=$(git diff --name-status HEAD~1...HEAD .)
+CHANGED_FILES=$(git diff --name-status HEAD~1...HEAD "$IMAGE_NAME-docker")
+
+
+if [ "$IMAGE_NAME" == "theia" ]
+then
+    if [ -z "$CHANGED_FILES" ]
+    then
+        # there were no changes in theia-docker
+        
+        # we want to build the theia-docker image also
+        # in the case that there were changes in other
+        # non theia*-docker folders
+        # eg. build scripts
+    
+        CHANGED_FILES_NON_DOCKER=$(git diff --name-status HEAD~1...HEAD .)
+        while read -r line; do
+            # the output of the git diff is of the form "M    theia-somename-docker/xyz"
+            if [[ $line =~ ^[[:space:]]*.[[:space:]]+theia(.)*-docker(.)+ ]]; then
+                # there was a change to an theia*-docker folder (not theia-docker)
+                # this will be picked up by another build
+                :
+            else
+                # there was a change in a non theia*-docker folder
+                # build theia-docker to be sure no utility scripts are broken
+                CHANGED_FILES="$line"$'\n'"$CHANGED_FILES"
+            fi
+        done <<< "$CHANGED_FILES_NON_DOCKER"
+
+    fi
+fi
+
+
 if [ -z "$CHANGED_FILES" ]
 then
     # nothing changed, skip building


vince-fugnitto:
@marcdumais-work @lmcbout 

I made the suggestion that the `theia-full-docker` image be built and tested when no images are explicitly updated during a PR. Should we however default to checking all images? (to be on the safe side). I see two main options:

1. always build all images if **no** images contain updates (might be expensive - we lose the functionality of only testing images that have changes).
2. test only `theia-full-docker` if **no** images contain updates, and restart CI jobs manually if we believe they should be tested (defeats the purpose of automating the CI job)

WDYT?

lmcbout:
Solution 1: build all the times, every PR building everything: too expensive 

I think testing the theia-full-docker image should be enough, (Solution 2) . When a **PR** is merged, no need to test all images in the case where there is no changes affecting all images. 

We should have an option to compile selected images without any changes in that image upon a user request. May be we could pass a compilation flag that specify we want to build this/all image(s) with/without changes

May be another alternative is to build all **NEXT** images on a cron job at a specific time of the day every day !

What do you think?

vince-fugnitto:
> We should have an option to compile selected images without any changes in that image upon a user request. May be we could pass a compilation flag that specify we want to build this/all image(s) with/without changes

This seems like overkill to me (and not commonly used). We can simply restart builds from the travis page manually (for example if I want to run `theia-cpp-docker`).

> May be another alternative is to build all **NEXT** images on a cron job at a specific time of the day every day !

I believe this is already the case.

lmcbout:
>>We should have an option to compile selected images without any changes in that image upon a >>user request. May be we could pass a compilation flag that specify we want to build this/all >>image(s) with/without changes

>This seems like overkill to me (and not commonly used). We can simply restart builds from the travis >page manually (for example if I want to run theia-cpp-docker).

I think if you start the build from travis with no changes in the image,  with the latest changes done for issue [236,](https://github.com/theia-ide/theia-apps/pull/236) all images will skip building

>>May be another alternative is to build all NEXT images on a cron job at a specific time of the day >>every day !

>I believe this is already the case.

If this is the case , then perfect.

vince-fugnitto:
> I think if you start the build from travis with no changes in the image, with the latest changes done for issue [236,](https://github.com/theia-ide/theia-apps/pull/236) all images will skip building

This PR builds the `theia-full-docker` in the following case. I believe it's enough, but just wanted feedback if we should build `theia-full-docker` or all images.

lmcbout:
Good way to push your PR before mine  :)

vince-fugnitto:
> Good way to push your PR before mine :)

I'm sorry, which PR?

dwjbosman:
> I think if you start the build from travis with no changes in the image, with the latest changes done for issue [236,](https://github.com/theia-ide/theia-apps/pull/236) all images will skip building

#236 was adapted in such a way that it will build all images when the event type is not "pull_request". 

TRAVIS_EVENT_TYPE: 
- push -> all images tested **(is the correct behavior?)**
- pull_request -> 
  -  only specific image tested
  -  (#247) if the change was not for a specific image, theia-full-docker is tested
- api -> (eg. via travis  dashboard) -> all images tested
- cron -> all images tested

dwjbosman:
> @dwjbosman I've spoken to @marcdumais-work and the `full` image might not be the best one to test since it's quite long and is the most fragile due to `gpg` keys.
> 
> Can we update the PR to instead use the [`theia-docker`](https://github.com/theia-ide/theia-apps/tree/master/theia-docker) image as the default case?

Changed it. Turned out there was a mistake in the regular expression as well.

check_changed.sh
```diff
@@ -16,9 +16,39 @@ then
...
+        CHANGED_FILES_NON_DOCKER=$(git diff --name-status HEAD~1...HEAD .)
+        while read -r line; do
+            # the output of the git diff is of the form "M    theia-somename-docker/xyz"
+            if [[ $line =~ ^[[:space:]]*.[[:space:]]+theia(.)*-docker(.)+ ]]; then
+                # there was a change to an theia*-docker folder (not theia-docker)
+                # this will be picked up by another build
+                :
+            else
+                # there was a change in a non theia*-docker folder
+                # build theia-full-docker to be sure no utility scripts are broken
```
vince-fugnitto: The comment should be updated as well correct?
`# build theia-docker ...`

The same is true for the PR title.

vince-fugnitto:
Thank you for making the updates @dwjbosman 👍

Issue #253, opened 2019-10-18
[rust] next theia-rust images have been failing CI recently


vince-fugnitto:
**Description**

The next `theia-rust` images have recently been failing during CI (cron) due to the following error:

```
error[E0407]: method `mixed_site` is not a member of trait `server::Span`
   --> /root/rust/cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-syntax-606.0.0/ext/proc_macro_server.rs:669:5
    |
669 | /     fn mixed_site(&mut self) -> Self::Span {
670 | |         self.mixed_site
671 | |     }
    | |_____^ not a member of trait `server::Span`
error: aborting due to previous error
For more information about this error, try `rustc --explain E0407`.
error: failed to compile `racer v2.1.28`, intermediate artifacts can be found at `/tmp/cargo-install4fnaaL`
Caused by:
  Could not compile `rustc-ap-syntax`.
To learn more, run the command again with --verbose.
The command '/bin/sh -c . /root/rust/cargo/env && umask 0002 && cargo +$rust_channel install racer' returned a non-zero code: 101
The command "./build_container.sh $NPM_TAG $IMAGE_NAME $NODE_VERSION" failed and exited with 101 during .
```

Only the `theia-rust:next` image has been failing during CI.
cc @dwjbosman 

**Additional Information**

_Last 3 Cron Builds:_
- [dashboard](https://travis-ci.org/theia-ide/theia-apps/builds)
- https://travis-ci.org/theia-ide/theia-apps/jobs/599435330
- https://travis-ci.org/theia-ide/theia-apps/jobs/598951372
- https://travis-ci.org/theia-ide/theia-apps/jobs/597933864

dwjbosman:
I will have a look.
   On Friday, October 18, 2019, 2:33:45 PM GMT+2, Vincent Fugnitto <notifications@github.com> wrote:  
 
 
Description

The next theia-rust images have recently been failing during CI (cron) due to the following error:
error[E0407]: method `mixed_site` is not a member of trait `server::Span`
   --> /root/rust/cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-syntax-606.0.0/ext/proc_macro_server.rs:669:5
    |
669 | /     fn mixed_site(&mut self) -> Self::Span {
670 | |         self.mixed_site
671 | |     }
    | |_____^ not a member of trait `server::Span`
error: aborting due to previous error
[...]

PR #254, opened 2019-10-24
Fix for #253 - Rust Channel update

dwjbosman:
Fix for #253 

Channel date has been updated.

### DIFF
--- a/theia-rust-docker/Dockerfile
+++ b/theia-rust-docker/Dockerfile
@@ -40,7 +40,7 @@ RUN mkdir -p $RUST_HOME && \
     chmod g+rwxs $RUST_HOME 
 RUN umask 0002 && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > /tmp/sh.rustup.rs && sh /tmp/sh.rustup.rs -y
 
-ARG rust_channel=nightly-2019-08-01
+ARG rust_channel=nightly-2019-11-25
 
 # install language server dependencies for theia
 RUN . /root/rust/cargo/env && umask 0002 && rustup toolchain install $rust_channel
--- a/theia-rust-docker/makefile
+++ b/theia-rust-docker/makefile
@@ -9,11 +9,11 @@ DOCKER_CONTAINER = theia_rust_container
 BUILD_DIR = build-env
 
 ifeq ($(RUST_CHANNEL),)
-RUST_CHANNEL := nightly-2019-08-25
+RUST_CHANNEL := nightly-2019-11-25
 endif
 
 # build the image using the cache
-image: 
+image:
     docker build --build-arg rust_channel=$(RUST_CHANNEL) -t $(DOCKER_IMAGE) .
 
 # This empty directory '$(BUILD_DIR)' exists in git, with only an .gitignore file, set to ignore the entire contents of the dir.


vince-fugnitto:
@marcdumais-work the CI is still failing at the same point, are we sure that it is the proper fix?

https://travis-ci.org/theia-ide/theia-apps/jobs/602497745#L1707-L1723

dwjbosman:
> @marcdumais-work the CI is still failing at the same point, are we sure that it is the proper fix?
> 
> https://travis-ci.org/theia-ide/theia-apps/jobs/602497745#L1707-L1723

Oh, I see.

I updated the channel argument in the Makefile, but not the default value in the Dockerfile (which is used by travis).

I'll update that as well

vince-fugnitto:
Unfortunately the CI is still failing for `theia-rust-docker:next` :(

https://travis-ci.org/theia-ide/theia-apps/jobs/603027093#L1535-L1543

dwjbosman:
Now Travis log contains  an "operation not permitted".

However if I build the image locally it succeeds.

The build_container.sh calls docker build. I ran this locally:

In theia-rust-docker:

`docker build --build-arg "version=next" --build-arg "NODE_VERSION=10.15.3"  -t theia-rust-docker `

Can anyone else try this

marcdumais-work:
> Can anyone else try this?

Trying now

marcdumais-work:
> Trying now

It built locally for me as well, using the same `docker build` command as above.

marcdumais-work:
@dwjbosman I tried a few things to troubleshoot the issue and I am left clueless at to why we have this result. 

Searching around for the error, it seems likely this is permission-related, but I do not understand why this happens on Travis but not locally. Maybe some Docker weirdness is at play.

dwjbosman:
I will do some online research later

vince-fugnitto:
I've updated the PR to include the following update to the Travis config file.
I've added the config property for `sudo: required` which fixes issues regarding operations which are not permitted. The same property is present in the main repository. Hopefully it will fix the issue.

I've also taken the liberty of squashing the PR so that the `DCO` check successfully passes.

vince-fugnitto:
Unfortunately it doesn't look like adding `sudo: required` works correctly.
I've seen some issues regarding the use of `cargo` and `rustup` and how permissions must be updated.

vince-fugnitto:
@dwjbosman have you gotten a chance to re-try to see how to resolve the error, the image is still failing CI.

vince-fugnitto:
@dwjbosman great ! the CI is now passing again :)
The only thing to do would be to clean up the history so it can be merged.

dwjbosman:
Another issue is that even though I pin the rust version to a specific channel (nightly-<date>) after some time the image will not build anymore. It seems as if some Rust components are not tied to the specific channel. 

    On Friday, November 29, 2019, 12:54:06 AM GMT+1, Vincent Fugnitto <notifications@github.com> wrote:  
 
 
@dwjbosman great ! the CI is now passing again :)
The only thing to do would be to clean up the history so it can be merged.

marcdumais-work:
> Another issue is that even though I pin the rust version to a specific channel (nightly-<date>) after some time the image will not build anymore. It seems as if some Rust components are not tied to the specific channel. 

Hi @dwjbosman ,

One potential approach would be to split this image in two: have a base image that has the Rust tools and as-little-else-as-possible (minimising the need to rebuild it for reasons other than updating the Rust tools), and then have a second image that extends the latest version of the first one to add the Theia application and everything else that's needed. This way we would have control when to update the Rust tools and rebuild/publish a new version of the first image.  Between updates, we would keep re-using the existing version of the first image that's known to work well.

Thinking of it, it could be even simpler: have our CI re-build the first image nightly, using e.g. the most recent version of Rust tools (nightly channel or such?), and whenever it builds, publish it as e.g. `theiaide/theia-rust-image-base:latest`. The second image can start from that (`FROM theiaide/theia-rust-image-base:latest`), which will automatically pick-up the last one that built. This would require tweaks in how we do CI / test the images, unless you setup the first image to be built / published elsewhere and we just consume it from this repo here. (there is precedent for doing this type of thing: https://github.com/theia-ide/theia-apps/blob/master/theia-swift-docker/Dockerfile#L28)

WDTY?

marcdumais-work:
I have just confirmed that this PR still passes CI. Even though the longer-term strategy is not yet determined, should we merge this for right now? 

@dwjbosman if you have the time, could you rebase this PR and squash the commits? Else maybe @vince-fugnitto can take care of this, preserving your authorship of the resulting commit.

vince-fugnitto:
@marcdumais-work I've rebased it, we can wait for the CI to pass before merging :)

vince-fugnitto:
@marcdumais-work the CI now passes successfully for **rust**, shall we merge?

marcdumais-work:
> @marcdumais-work the CI now passes successfully for **rust**, shall we merge?

Yes, please do.

PR #258, opened 2019-11-02
Add 'cpp-debug' dependency

### LINKED ISSUE #227: Update references of 'eclipse/theia-cpp-extension'
The repo **eclipse/theia-cpp-extension** has been moved and renamed to
 [**eclipse-theia/theia-cpp-extensions**](https://github.com/eclipse-theia/theia-cpp-extensions). We need to adjust any reference to the old org and name.

_Example_:

https://github.com/theia-ide/theia-apps/blob/9338e6498d8498719466e125e0108f95033f69ae/theia-rust-docker/Dockerfile#L23-L24

cc @dwjbosman

vince-fugnitto:
Fixes #227

- added `@theia/cpp-debug` dependency to the `theia-cpp-docker:latest` image.
- added `@theia/cpp-debug` dependency to the `theia-cpp-docker:next` image.
- added `@theia/cpp-debug` dependency to the `theia-cpp-electron` application.
- updated documentation present in the `theia-cpp-docker` readme.
- updated documentation present in the `theia-cpp-electron` readme.

### DIFF
--- a/theia-cpp-docker/README.md
+++ b/theia-cpp-docker/README.md
@@ -8,7 +8,7 @@ A containerized Theia-based C/C++ demo IDE, including commonly used tools:
 
 The included Theia-based IDE application has the following notable features
 - [@theia/cpp] Language-server built-in clang-tidy static analyser integration. Will analyse files opened in the IDE's editors and report problems for configured rules. See [README](https://github.com/theia-ide/theia/tree/master/packages/cpp#using-the-clang-tidy-linter) for more details, including related preferences
-- [TODO][@theia/cpp-debug] basic C/C++ debugging support
+- [@theia/cpp-debug] Basic C/C++ debugging support
 
 ### How to use
 
@@ -24,7 +24,7 @@ Options:
 
 ### How to build
 
-Build image using `next` Theia packages and strip the Theia application to save space (with reduced debuggability) 
+Build image using `next` Theia packages and strip the Theia application to save space (with reduced debuggability)
 ```bash
 docker build --no-cache --build-arg version=next --build-arg strip=true  -t theia-cpp:next .
 ```
--- a/theia-cpp-docker/latest.package.json
+++ b/theia-cpp-docker/latest.package.json
@@ -18,6 +18,7 @@
         "@theia/console": "latest",
         "@theia/core": "latest",
         "@theia/cpp": "latest",
+        "@theia/cpp-debug": "latest",
         "@theia/debug": "latest",
         "@theia/debug-nodejs": "latest",
         "@theia/editor": "latest",
--- a/theia-cpp-docker/next.package.json
+++ b/theia-cpp-docker/next.package.json
@@ -18,6 +18,7 @@
         "@theia/console": "next",
         "@theia/core": "next",
         "@theia/cpp": "next",
+        "@theia/cpp-debug": "next",
         "@theia/debug": "next",
         "@theia/debug-nodejs": "next",
         "@theia/editor": "next",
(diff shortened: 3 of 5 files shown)


vince-fugnitto:
@lmcbout 

> you need to add the resolution to each package.json (latest & next) to allow saving :
> 
> ```
>     "resolutions": {     
>         "vscode-languageserver-protocol": "3.15.0-next.9",
>         "vscode-languageserver-types": "3.15.0-next.5"
>       },
> ```

I don't think that resolution should be added in this PR.
It is already being tracked with #260 and I believethe proper fix is to patch upstream.

> Question 1:
> I have in my preference setting "cpp.clangdExecutable": "clangd-10",
> but it seems the preference is not being looked at anymore (Probably not related to this PR ), so it does not report the clang-tidy info

Does the preference exist in the `preferences-tree`?

> Question2:
> this PR is taking care of issue #227 (theia-rust-docker). Do you need to add in theia-rust-docker "@theia-cpp" with "@theia-cpp-debug " to work ?

No, the `@theia/cpp-debug` is used in theia-rust-docker since it uses `gdb` for the debug console.
There is no need for additional C/C++ features.

lmcbout:
>I don't think that resolution should be added in this PR.
It is already being tracked with #260 and I believe the proper fix is to patch upstream.

If it is tracked in #260, then #260 has to be pushed before your PR since this configuration is useless, no one can save any data.

>Does the preference exist in the preferences-tree?

Yes it does, but as said, It is probably a new issue we should track, it does not depend on this PR

About the theia-rust-docker  image, thanks for the info, I was not sure

vince-fugnitto:
> > I don't think that resolution should be added in this PR.
> > It is already being tracked with #260 and I believe the proper fix is to patch upstream.
> 
> If it is tracked in #260, then #260 has to be pushed before your PR since this configuration is useless, no one can save any data.

Useless is a strong word, I'm not adding new configurations, I'm adding new dependencies for `@theia/cpp-debug` which brings it's own functionalities. I don't see the configs you are referring to.

vince-fugnitto:
@lmcbout which version are you referring to?

lmcbout:
>which version are you referring to?

For example: theia-apps/theia-cpp-docker/latest.package.json
At the beginning of the file, you see the field "version" which is "0.0.1", this version should be updated to the next one. Result: When you run Theia -> Help -> Getting started", you will see  at the bottom the new version being displayed. The reason to change it is because you added new functionality and it would be wise to know on which version it was added.

vince-fugnitto:
> > which version are you referring to?
> 
> For example: theia-apps/theia-cpp-docker/latest.package.json
> At the beginning of the file, you see the field "version" which is "0.0.1", this version should be updated to the next one. Result: When you run Theia -> Help -> Getting started", you will see at the bottom the new version being displayed. The reason to change it is because you added new functionality and it would be wise to know on which version it was added.

@lmcbout these packages have never been bumped and the packages never published, why start updating their versions now with this PR (when there has been many previous updates to functionality)?

vince-fugnitto:
@lmcbout if you check, there are some docker images without any versions as well.

lmcbout:
@vince-fugnitto  I think this would be nice for future reference, know when a feature was added, easier to know the end-user configuration, but at the end, this is **not mandatory**.

If you notice, I already approved this PR and I put "but this is not necessary"

vince-fugnitto:
> @vince-fugnitto I think this would be nice for future reference, know when a feature was added, easier to know the end-user configuration, but at the end, this is **not mandatory**.
> 
> If you notice, I already approved this PR and I put "but this is not necessary"

Yes just don't see why it was mentioned in **this** PR. I think before we start bumping versions and everything we should align all images, and think of a strategy for the future of when and why to update. I don't just want to start on a random pull request and update only a single image when others can't be updated with a similar strategy.

Issue #65, opened 2018-07-14
[eslint] add 'eslint' support for ts, js images


bdefore:
Would be very helpful for JavaScript development to have inline eslint warnings in Theia, along the lines of what we currently have for Typescript.

FWIW, this is something currently supported by https://codesandbox.io/

akosyakov:
We should make use of eslint LS: https://github.com/Microsoft/vscode-eslint/tree/master/server

sr229:
Not sure if Theia's LS is VSCode compatible, we might need to rewrite `vscode-eslint` for Theia

marcdumais-work:
update: since last fall Theia has increasingly gained compatibility with VS Code extensions (run as Theia plugins). I do not remember if `eslint` was tested, but there's a fair chance it would work out-of-the-box.

realies:
Can ESLint be ran in the Theia docker container somehow? Setting the plugin path as an environment variable and placing the vsix in the folder does not seem to work.

marcdumais-work:
@realies yes, I just confirmed that it works. For convenience I started with an image that already has plugin support: theia-go, and added dependencies to support TypeScript (@theia/typescript, typescript), installed eslint globally and added the .vsix file in the plugins folder. 

See this for reference: https://github.com/marcdumais-work/theia-apps/commit/ebf1ef62dd92fa8c9b54a6a2f7639d5f8e6aecd6

I tested using this repo: https://github.com/trevordmiller/example-eslint and got eslint diagnostics and the eslint commands are available from the command palette.

![image](https://user-images.githubusercontent.com/25749063/62213111-35183a00-b370-11e9-8cdc-45a066e536b3.png)

vince-fugnitto:
> @realies yes, I just confirmed that it works. For convenience I started with an image that already has plugin support: theia-go, and added dependencies to support TypeScript (@theia/typescript, typescript), installed eslint globally and added the .vsix file in the plugins folder.
> 
> See this for reference: [marcdumais-work@ebf1ef6](https://github.com/marcdumais-work/theia-apps/commit/ebf1ef62dd92fa8c9b54a6a2f7639d5f8e6aecd6)
> 
> I tested using this repo: https://github.com/trevordmiller/example-eslint and got eslint diagnostics and the eslint commands are available from the command palette.
> 
> ![image](https://user-images.githubusercontent.com/25749063/62213111-35183a00-b370-11e9-8cdc-45a066e536b3.png)

[...]

PR #259, opened 2019-11-02
Update 'stalebot' to only check 'question' labels

vince-fugnitto:
**Description**

Update the `stalebot` config to only verify issues which have the `question` label. Now that https://github.com/probot/stale/pull/213 is fixed, we can turn stalebot on with the initial plan of only
verifying issues which have a certain type of label.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -24,7 +24,7 @@ daysUntilStale: 60
 daysUntilClose: 30
 
 # Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
-onlyLabels: []
+onlyLabels: ['question']
 
 # Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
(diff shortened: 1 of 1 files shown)


vince-fugnitto:
Motivation for the pull-request: https://github.com/theia-ide/theia-apps/issues/65#issuecomment-549010656

marcdumais-work:
I propose we hold-off a bit and instead start to systematically apply the `no_stale` label whenever we think an issue is good to keep around even if it may not be addressed in the short term. 

I just did a quick pass for existing issues and PRs.

vince-fugnitto:
> I propose we hold-off a bit and instead start to systematically apply the `no_stale` label whenever we think an issue is good to keep around even if it may not be addressed in the short term.
> 
> I just did a quick pass for existing issues and PRs.

I agree, I think the bot itself has been useful in closing old our outdated issues and pull requests.
I think we can manually use the `no-stale` label whenever we judge that an issue should always be applicable.

PR #262, opened 2019-11-07
fix issue #260 - editor can no longer be saved

### LINKED ISSUE #260: [all images] Can't save from IDE
There is a recent breaking change caused by some change in one or more Theia NPM dependencies, that causes the save feature not to work.

The problem is still being investigated. See: https://github.com/eclipse-theia/theia/issues/6500

lmcbout:
Signed-off-by:Jacques Bouthillier <jacques.bouthillier@ericsson.com>

**Reference issue**
Fix #260 

**What it does**
It allows to save the text modification after editing.

**How to test**
Open any Theia-app
Open a file in the editor 
Make any modification  
Save it

**Review checklist**
- [x] as an author, I have thoroughly tested my changes and carefully followed the review guidelines

**Reminder for reviewers**
- [ ] as a reviewer, I agree to behave in accordance with the review guidelines

### DIFF
--- a/theia-cpp-docker/latest.package.json
+++ b/theia-cpp-docker/latest.package.json
@@ -59,6 +59,10 @@
         "@theia/workspace": "latest",
         "typescript": "latest"
     },
+    "resolutions": {
+      "vscode-languageserver-protocol": "3.15.0-next.9",
+      "vscode-languageserver-types": "3.15.0-next.5"
+    },
     "devDependencies": {
         "@theia/cli": "latest"
     }
--- a/theia-cpp-docker/next.package.json
+++ b/theia-cpp-docker/next.package.json
@@ -59,6 +59,10 @@
         "@theia/workspace": "next",
         "typescript": "latest"
     },
+    "resolutions": {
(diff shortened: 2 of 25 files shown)


vince-fugnitto:
@lmcbout as I mentioned, the issue should be resolved upstream (@marechal-p is looking into a fix), so there is no need to update all images.

vince-fugnitto:
@lmcbout is it correct that the PR contains two commits?
 (I believe because it may be based off of #258)

lmcbout:
@vince-fugnitto sorry, I did not mean to include your PR with mine, it should have been just one commit and I just did

lmcbout:
@marechal-p  I agree with you, we will revert this PR when the main issue is solved. In the mean time, we can use the theia-app.

marcdumais-work:
> I was tentatively looking at it in my yarn upgrade PR, but I have a hard time figuring things out.
> 
> This could be a good temporary fix to revert once everything is back in order.

Sounds like a plan - I would not mind waiting a bit for an upstream fix if our CI was catching the problem and preventing publishing of flawed images, but it's not.

PR #265, opened 2019-11-08
[full] Define THEIA_DEFAULT_PLUGINS, fix build

### LINKED ISSUE #263: [full] Setup THEIA_DEFAULT_PLUGINS so we can use plugins
Hi! I wondered if I can install a VSCode extension on theia-full.
I've tried to install liveshare extension but i can't, because the plugins folder where I can put .vsix  doesn't exist, but considering that we are in a container the folder persistence is low.  

What is the right way to install a VSCode extension?

marcdumais-work:
So that we can use Theia plugins / VS Code extensions in this image.

Also fixes full image build.

Fixes #263

### DIFF
--- a/theia-full-docker/Dockerfile
+++ b/theia-full-docker/Dockerfile
@@ -98,7 +98,8 @@ RUN curl -sS https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar
     go get -u -v github.com/josharian/impl && \
     go get -u -v github.com/tylerb/gotype-live && \
     go get -u -v github.com/rogpeppe/godef && \
-    go get -u -v golang.org/x/tools/cmd/godoc && \
+    # godoc causing issues ATM, preventing image to build - commenting-out
+    # go get -u -v golang.org/x/tools/cmd/godoc && \
     go get -u -v github.com/zmb3/gogetdoc && \
     go get -u -v golang.org/x/tools/cmd/goimports && \
     go get -u -v sourcegraph.com/sqs/goreturns && \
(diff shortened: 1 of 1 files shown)


theia-full-docker/Dockerfile
```diff
@@ -181,7 +182,8 @@ ARG GITHUB_TOKEN
 RUN yarn --cache-folder ./ycache && rm -rf ./ycache && \
     NODE_OPTIONS="--max_old_space_size=4096" yarn theia build
 EXPOSE 3000
-ENV SHELL /bin/bash
+ENV SHELL /bin/bash\
+    THEIA_DEFAULT_PLUGINS=local-dir:/home/theia/plugins
```
marechal-p: This set the `SHELL` var to `/bin/bash THEIA_DEFAULT_PLUGINS=local-dir:/home/theia/plugins`, I think this is a typo and you wanted to define a second env var ?

marcdumais-work: Indeed! 

I think this can be fixed just by replacing the space between `SHELL` and `/bin/bash`  with an `=` , since `ENV` supports defining multiple in one command. 

see doc, towards the middle of the page: https://docs.docker.com/engine/reference/builder/

exerpt:
```
The second form, ENV <key>=<value> ..., allows for multiple variables to be set at one time. Notice that the second form uses the equals sign (=) in the syntax, while the first form does not. Like command line parsing, quotes and backslashes can be used to include spaces within values.

For example:

ENV myName="John Doe" myDog=Rex\ The\ Dog \
    myCat=fluffy
```

PR #277, opened 2019-12-05
Fix SHELL environment variable

marechal-p:
There was a typo, making the SHELL var corrupted.

### DIFF
--- a/theia-full-docker/Dockerfile
+++ b/theia-full-docker/Dockerfile
@@ -182,7 +182,7 @@ ARG GITHUB_TOKEN
 RUN yarn --cache-folder ./ycache && rm -rf ./ycache && \
     NODE_OPTIONS="--max_old_space_size=4096" yarn theia build
 EXPOSE 3000
-ENV SHELL /bin/bash\
+ENV SHELL=/bin/bash \
     THEIA_DEFAULT_PLUGINS=local-dir:/home/theia/plugins
 
 ENV GOPATH=/home/project


vince-fugnitto:
@marechal-p make sure to `sign-off` so the dco passes :)

theia-full-docker/Dockerfile
```diff
@@ -182,8 +182,8 @@ ARG GITHUB_TOKEN
 RUN yarn --cache-folder ./ycache && rm -rf ./ycache && \
     NODE_OPTIONS="--max_old_space_size=4096" yarn theia build
 EXPOSE 3000
-ENV SHELL /bin/bash\
-    THEIA_DEFAULT_PLUGINS=local-dir:/home/theia/plugins
+ENV SHELL /bin/bash
+ENV THEIA_DEFAULT_PLUGINS local-dir:/home/theia/plugins
```
marcdumais-work: This works much better. 

Consider using the [second form](https://github.com/theia-ide/theia-apps/pull/265#discussion_r354414132) for `ENV` that permits defining multiple key-value pairs and saves us from adding extra docker image layers.

marechal-p: Fixed

marechal-p:
I think something I did changed the GO related variables, building locally to double check.

marechal-p:
Yup, docker doesn't seem to evaluate cross references within the same ENV statement.

The following seemed to be failing:

```Dockerfile
ENV a=1 \
    b=$a:2
```

Reverted back to the two separate statements.

PR #281, opened 2019-12-18
docker client + docker-compose + docker-machine support

brandonros:
This stands no chance of getting accepted but I wanted to save anybody else going through this the headache of trying to piece this together from scratch.

### DIFF
--- a/theia-docker/Dockerfile
+++ b/theia-docker/Dockerfile
@@ -1,10 +1,13 @@
-ARG NODE_VERSION=10
-FROM node:${NODE_VERSION}-alpine
+# 0. Docker client
+FROM docker:latest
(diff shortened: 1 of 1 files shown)


vince-fugnitto:
> This stands no chance of getting accepted but I wanted to save anybody else going through this the headache of trying to piece this together from scratch.

Thank you for your contribution @brandonros. I'm curious, why did you close the pull request before proceeding with a review? I think it is definitely interesting to understand the reasoning behind the pull request and adding support for `docker client`, `docker-compose` and `docker-machine`. If you are worried about it not being accepted, it is always possible to submit a contribution which adds a new image which satisfies the functionality requirements you described.

PR #284, opened 2020-01-13
wip: [vscode] migrate to builtin extensions

vince-fugnitto:
#### What it does

The idea is to update the `theia-full` image with the necessary VS Code builtins and remove soon to be deprecated language specific extensions. Any missing functionality should be covered by external VS Code extensions.

#### How to test

TBD.

#### Review checklist

- [ ] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers


theia-full-docker/next.package.json
```diff
@@ -68,5 +57,78 @@
...
+    },
+    "adapterDir": "plugins",
+    "adapters": {
+        "vscode-builtin-bat": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/bat-1.39.1-prel.vsix",
+        "vscode-builtin-clojure": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/clojure-1.39.1-prel.vsix",
+        "vscode-builtin-coffeescript": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/coffeescript-1.39.1-prel.vsix",
+        "vscode-builtin-cpp": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/cpp-1.39.1-prel.vsix",
+        "vscode-builtin-csharp": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/csharp-1.39.1-prel.vsix",
+        "vscode-builtin-css": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/css-1.39.1-prel.vsix",
+        "vscode-builtin-css-language-features": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/css-language-features-1.39.1-prel.vsix",
```
akosyakov: we cannot use all built-in extension, please consult corresponding issue in Theia, it should explain why some extensions should not be used yet

akosyakov:
@marcdumais-work @vince-fugnitto Could we first migrate theia repo? It is more important, after that theia apps can be migrated.

vince-fugnitto:
> @marcdumais-work @vince-fugnitto Could we first migrate theia repo? It is more important, after that theia apps can be migrated.

@akosyakov I’m fine with updating the main repo first. Would the idea be to remove language specific extensions from the examples only, or also remove them from the codebase and into their own repos (under `theiaide`)?

akosyakov:
>  I’m fine with updating the main repo first. Would the idea be to remove language specific extensions from the examples only, or also remove them from the codebase and into their own repos (under theiaide)?

I was thinking about removing them completely. If no one using them why they should be maintained? We should push it in the right direction by using VS Code built-in extensions ourself, other will follow eventually. Also code still in git if we decide to reverse this decision. Please also delete the extension manage extension it is not used for sure.

vince-fugnitto:
> > I’m fine with updating the main repo first. Would the idea be to remove language specific extensions from the examples only, or also remove them from the codebase and into their own repos (under theiaide)?
> 
> I was thinking about removing them completely. If no one using them why they should be maintained? We should push it in the right direction by using VS Code built-in extensions ourself, other will follow eventually. Also code still in git if we decide to reverse this decision. Please also delete the extension manage extension it is not used for sure.



> > I’m fine with updating the main repo first. Would the idea be to remove language specific extensions from the examples only, or also remove them from the codebase and into their own repos (under theiaide)?
> 
> I was thinking about removing them completely. If no one using them why they should be maintained? We should push it in the right direction by using VS Code built-in extensions ourself, other will follow eventually. Also code still in git if we decide to reverse this decision. Please also delete the extension manage extension it is not used for sure.

I understand, I was just thinking for the sake of our images and also if a product currently references the language extensions. The next release will likely break if we do not publish them. I think at the very least we should advertise that such extensions should no longer be used to prepare them for the migration as well.

We should also refactor current extensions which have direct dependencies on languages (I think `@theia/json` is one such example).

akosyakov:
> I understand, I was just thinking for the sake of our images and also if a product currently references the language extensions. The next release will likely break if we do not publish them. I think at the very least we should advertise that such extensions should no longer be used to prepare them for the migration as well.

Ok, let's for now remove them only from examples package.json and remove when theia-apps are migrated. I would be fine with anything which allow to switch theia examples. I think we should do it in first place before upgrading downstream clients.

vince-fugnitto:
> > I understand, I was just thinking for the sake of our images and also if a product currently references the language extensions. The next release will likely break if we do not publish them. I think at the very least we should advertise that such extensions should no longer be used to prepare them for the migration as well.
> 
> Ok, let's for now remove them only from examples package.json and remove when theia-apps are migrated. I would be fine with anything which allow to switch theia examples. I think we should do it in first place before upgrading downstream clients. Otherwise we keep developing and testing against old Theia extensions which should not be actually used.

I'll make the changes :) and I agree, the earlier we start testing against the builtins the better.

PR #287, opened 2020-01-21
[java-docker] use builtin and external ext

vince-fugnitto:
#### What it does

Update the `theia-java-docker:next` image to include builtins and external VS Code extensions

#### How to test
Build and verify the changes (that plugins are downloaded and picked up when started)

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers


akosyakov:
@marcdumais-work @vince-fugnitto Would be fine If i do `theia-ide/theia`?

One important thing is Electron. It is not clear how to bundle extensions yet. An end user cannot use env variables or cli options to pick up extensions, they should be inside the bundled app and picked up automatically.

vince-fugnitto:
> @marcdumais-work @vince-fugnitto Would be fine If i do `theia-ide/theia`?

Please go ahead :) I wanted to try it out for an example image (java) so I could iron out any details and see what needs to be done for other images.

> One important thing is Electron. It is not clear how to bundle extensions yet. An end user cannot use env variables or cli options to pick up extensions, they should be inside the bundled app and picked up automatically.

 I believe @lmcbout had some experience picking up plugins for Electron applications if we get blocked.

akosyakov:
> I believe @lmcbout had some experience picking up plugins for Electron applications if we get blocked.

I've seen the issue on Theia repo where it was recommended an env variable, but for builtins it is not how it should work. A user should download distributable, run it and VS Code extensions should be there already, without changing any env variables.

marcdumais-work:
> > I believe @lmcbout had some experience picking up plugins for Electron applications if we get blocked.
> 
> I've seen the issue on Theia repo where it was recommended to configure an env variable, but for builtins it is not how it should work. A user should download distributable, run it and VS Code extensions should be there already, without changing any env variables. I have not seen anyone to figure it out and it will be important for all Electron based products if we remove textmate grammars Theia extension.

In our current internal application we do have the built-ins and some extra VS Code extensions part of the bundle (MSI installer for Window, AppImage for Linux). We have our own version of `electron-main.js` that's used as entry-point, and where we inject env var `THEIA_DEFAULT_PLUGINS` pointing it to the bundled plugins folder. 

So it does rely on the env var mechanism but the user needs not set it or care about it. This leaves env var `THEIA_PLUGINS` to point to a folder on the host, where the user or eventually our extensions manager could install their own extensions.

akosyakov:
> We have our own version of electron-main.js that's used as entry-point, and where we inject env var THEIA_DEFAULT_PLUGINS pointing it to the bundled plugins folder.

Does not sound super clean :) Can we generalize it in Theia to a solution without using custom electron-main.js?

marcdumais-work:
> Does not sound super clean :) Can we generalize it in Theia to a solution without using custom electron-main.js?

Sure. Maybe the plugins system could look-for a plugins folder at a given path relative to the app, that would be used if present, without depending on any env var? Or maybe you have other ideas. 

For reference, he're how we do it ATM in our custom `electron-main.js`:
``` javascript
process.env.THEIA_DEFAULT_PLUGINS = `local-dir:${resolve(__dirname, '..', '..', '..', '..', 'plugins')}`;
```

akosyakov:
> Maybe the plugins system could look-for a plugins folder at a given path relative to the app, that would be used if present, without depending on any env var?

No ideas, i don't have much experience with Electron. We can try whatever you propose. Maybe someone come up with a better idea on the PR.

akosyakov:
Let's move a discussion here: https://github.com/eclipse-theia/theia/issues/6946

vince-fugnitto:
Unfortunately, we cannot successfully update the applications until `@theia/cli` is released.
We cannot include `@theia/cli:next` in the `latest.package.json` since it pulls `@theia/application-package` (used by many extensions) and will cause a multi injector error on startup.

We either need to:
- release a hotfix with the update `@theia/cli` so we can update the images ahead of the release (gives us time to update the images, gives others extra time to adopt their applications to the builtins)
- wait a week for the release and proceed with the general updates (with many other things to update before we can deprecate extensions anyway, maybe this approach is fine)

akosyakov:
@vince-fugnitto yes, it won't work, but we can upgrade next images for now? also i've noticed that latest image does not have plugins. I think we should include them in each image, since we build them separate docker containers, also it will make more complete standalone examples.

theia-java-docker/next.package.json
```diff
@@ -15,20 +15,89 @@
         "@theia/file-search": "next",
         "@theia/getting-started": "next",
         "@theia/git": "next",
-        "@theia/java": "next",
         "@theia/json": "next",
```
marcdumais-work: I think maybe we're keeping this one because the json built-in is not working yet?

theia-java-docker/next.package.json
```diff
@@ -15,20 +15,89 @@
...
         "@theia/preferences": "next",
+        "@theia/plugin": "next",
+        "@theia/plugin-ext": "next",
+        "@theia/plugin-ext-vscode": "next",
         "@theia/search-in-workspace": "next",
         "@theia/terminal": "next",
-        "@theia/textmate-grammars": "next",
-        "@theia/tslint": "next",
-        "@theia/typescript": "next",
         "typescript": "latest"
```
marcdumais-work: needed?

marcdumais-work:
Got these errors:

image


Maybe setting `JAVA_HOME` so it points to the JDK instead of the JRE would help:  "/opt/ibm/java/" ? (line 7 in `Dockerfile`)

marcdumais-work:
I tested a bit with a few sample Java projects. Once the JDK path issue is fixed, the image seems to work decently well. So looking good so far.

theia-java-docker/next.package.json
vince-fugnitto: Correct, the `vscode-builtin-json-language-features` is not yet working so I could not add it at this time (https://github.com/eclipse-theia/theia/issues/6623).

`vscode-builtin-json` is included to add JSON syntax highlighting and snippet support since `@theia/textmate` grammars is removed.

theia-java-docker/next.package.json
vince-fugnitto: > needed?

I wasn't sure, it looks like all images include `typescript` as a dependency.

theia-java-docker/latest.package.json
```diff
@@ -11,27 +11,97 @@
...
+        "vscode-java-dependency-viewer": "https://github.com/microsoft/vscode-java-dependency/releases/download/0.6.0/vscode-java-dependency-0.6.0.vsix"
     },
     "resolutions": {
         "vscode-languageserver-protocol": "3.15.0-next.9",
         "vscode-languageserver-types": "3.15.0-next.5",
         "**/vscode-json-languageserver/**/vscode-languageserver": "6.0.0-next.1"
     },
     "devDependencies": {
-        "@theia/cli": "latest"
+        "@theia/cli": "next"
```
akosyakov: latest

vince-fugnitto:
@marcdumais-work I've updated based on previous feedback, and since `@theia/cli` has been released last week both the `latest` and `next` images can benefit from the `download:plugins` script (used for example to pull builtin plugins).

marcdumais-work:
Merging

vince-fugnitto:
> @marcdumais-work @vince-fugnitto Would be fine If i do `theia-ide/theia`?

@akosyakov do you still plan on migrating the `theia-docker` image?

PR #297, opened 2020-02-06
python: update images to use vscode extensions

### LINKED ISSUE #74: Support of Python3

vince-fugnitto:
#### What it does

Fixes #74

The goal of the pull-request is to update the existing `theia-python:latest` and `theia-python:next` images to make use of both **builtin** and **external** vscode extensions while also removing soon-to-be-deprecated extensions such as `@theia/python`. 

The `Dockerfile` has also been modified to a modified version of the `theia-full-docker` image, complete with necessary Python 2 and 3 tools for development including the [python-language-server](https://pypi.org/project/python-language-server/) which supports both versions.

Python functionality is completed by the [vscode-python](https://marketplace.visualstudio.com/items?itemName=ms-python.python) extension which also supports both Python 2 and 3.

In essence, the pull-request includes:
- updates the `theia-python:latest` image to use builtin and vscode extensions.
- updates the `theia-python:next` image to use builtin and vscode extensions.
- updates the image to support both python 2 and 3.
- includes the necessary tools for python development.

#### How to test

Verify that the CI successfully passes for both `theia-python:latest` and `theia-python:next`.

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers


theia-python-docker/Dockerfile
```diff
@@ -1,22 +1,123 @@
...
+    B9AE9905FFD7803F25714661B63B535A4C206CA9 \
+    77984A986EBC2AA786BC0F66B01FBB92821C587A \
+    71DCFD284A79C3B38668286BC97EC7A07EDE3FC1 \
+    FD3A5288F042B6850C66B31F09FE44734EB7990E \
+    8FCCA13FEF1D0C2E91008E09770F7A9A5AE15600 \
+    C4F0DFFF4E8C1A8236409D08E73BC641CC11F4C8 \
+    DD8F2338BAE7501E3DD5AC78C273792F7D83545D \
+    A48C2BEE680E841632CD4E44F07496B3EB3C1762 \
+    ; do \
+    gpg --batch --keyserver ipv4.pool.sks-keyservers.net --recv-keys "$key" || \
```
lmcbout: Do we absolutely need to add the gpg keys, is teher another solution not depending on those keys?

theia-python-docker/latest.package.json
```diff
@@ -31,5 +33,72 @@
...
+        "vscode-builtin-typescript": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/typescript-1.39.1-prel.vsix",
+        "vscode-builtin-typescript-language-features": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/typescript-language-features-1.39.1-prel.vsix",
+        "vscode-builtin-vb": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/vb-1.39.1-prel.vsix",
+        "vscode-builtin-icon-theme-seti": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/vscode-theme-seti-1.39.1-prel.vsix",
+        "vscode-builtin-xml": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/xml-1.39.1-prel.vsix",
+        "vscode-builtin-yaml": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/yaml-1.39.1-prel.vsix",
+        "vscode-editorconfig": "https://github.com/theia-ide/editorconfig-vscode/releases/download/v0.14.4/EditorConfig-0.14.4.vsix",
+        "vscode-python": "https://github.com/microsoft/vscode-python/releases/download/2020.1.58038/ms-python-release.vsix"
```
lmcbout: any reason why we don't include the vscode-eslint part of the bundle ?

vince-fugnitto: Is there a reason you'd want to include `vscode-eslint` as part of a **Python** image?

theia-python-docker/Dockerfile
vince-fugnitto: This was copied over from the `theia-full-docker` image and I did not want to mess with the setup.

theia-python-docker/latest.package.json
lmcbout: Since the other builtin are included, for example the one concerning Typescript, why not including them all related to Typescript;
Already in the package.json
```
        "vscode-builtin-typescript": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/typescript-1.39.1-prel.vsix",
        "vscode-builtin-typescript-language-features": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/typescript-language-features-1.39.1-prel.vsix",
        "vscode-builtin-vb": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/vb-1.39.1-prel.vsix",
 
```

theia-python-docker/Dockerfile
lmcbout: Why not stating from another Dockersetup like java0docker !!!

theia-python-docker/latest.package.json
vince-fugnitto: @lmcbout the goal is to someday use all the `builtins` (since they are present in vscode), but for the moment only the known-to-work ones are included (this is why `typescript` is included).

`vscode-eslint` however is not a builtin, it is an external vscode extension.
I don't see how useful it'd be in a base Python image.

marcdumais-work: I assumed that "vscode-builtin-vb" was for Visual Basic?

theia-python-docker/Dockerfile
marcdumais-work: It looks like the [node:$NODE_VERSION](https://hub.docker.com/_/node/) docker images, weve been using in other Theia images of this repo, are Debian-based. For example image `node:10` is built from [this Dockerfile](https://github.com/nodejs/docker-node/blob/46f5203674c748b0701135c8eeea4b250b3ecb6d/10/stretch/Dockerfile)

Debian is close to Ubuntu (or the other way around), sharing the `.deb` package format. However the packages are not inter-operable. I think Ubuntu has the greater library of packages, specially with PPAs. 

So for our bigger images like `theia-full` I'm not sure we want to switch to the `node:$version` image as base (we could try as experiment). But this image here looks like it could work without too much trouble. @vince-fugnitto want to give it a try?

theia-python-docker/Dockerfile
```diff
@@ -1,22 +1,123 @@
...
+    && apt-get install -y python3-dev python3-pip \
     && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/*
 
-RUN pip install \
-    python-language-server \
-    flake8 \
-    autopep8
+## User account
+RUN adduser --disabled-password --gecos '' theia && \
+    adduser theia sudo && \
```
marcdumais-work: The previous version of this image I think did not have sudo capability? If so adding it could be considered a breaking change by some users, that may be running it in environment where this is not permitted. 

Any particular reason to include this now?

theia-python-docker/Dockerfile
```diff
@@ -1,22 +1,123 @@
...
+    echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers;
 
-ARG version=latest
+RUN chmod g+rw /home && \
+    mkdir -p /home/project && \
+    mkdir -p /usr/local/cargo && \
```
marcdumais-work: leftover?

marcdumais-work: as well as the following 5 lines? (golang / rust related folders I think)

theia-python-docker/Dockerfile
```diff
@@ -1,22 +1,123 @@
...
+RUN yarn --cache-folder ./ycache && rm -rf ./ycache && \
+    NODE_OPTIONS="--max_old_space_size=4096" yarn theia build && \
+    yarn theia download:plugins
+
 EXPOSE 3000
-ENV SHELL /bin/bash
+ENV SHELL=/bin/bash \
+    THEIA_DEFAULT_PLUGINS=local-dir:/home/theia/plugins
+
+ENV GOPATH=/home/project
```
marcdumais-work: leftover?

theia-python-docker/Dockerfile
```diff
@@ -1,22 +1,123 @@
...
+ENV PATH=$PATH:$GOPATH/bin
```
marcdumais-work: leftover?

theia-python-docker/Dockerfile
vince-fugnitto: I believe the only reason it is now included is since it was copied over from 
the `theia-full-docker` image. Do we expect to not be breaking in the images? 
(I thought they were for education purposes like examples on how to build their own images or for us to test functionality present in the main repo)

If it causes issues I can try to remove it and see if it still works correctly.

vince-fugnitto: I see the following comment:

> Git and sudo (sudo needed for user override)

https://github.com/theia-ide/theia-apps/pull/297/files#diff-db44fcad56aa1590715c9f3963aa5a86R77

theia-python-docker/latest.package.json
vince-fugnitto: > I assumed that "vscode-builtin-vb" was for Visual Basic?

Yes exactly, `vscode-builtin-vb` is for Visual Basic: https://github.com/microsoft/vscode/blob/master/extensions/vb/package.json#L16

theia-python-docker/Dockerfile
vince-fugnitto: I can definitely give it a try :)

marcdumais-work:
(Note: I tested with a version of this PR where I switched to the `node:10` image as base. I then built a `latest` version of the image. So take with a grain of salt - some issues may not reproduce on the original image)

the python extension logs a lot, specially at the `INFO` level. We should configure logging so these are not shown. 

There are a couple of runtimes exceptions in BE log from the `python` extension. Here's one that I think happened during load:

root ERROR [hosted-plugin: 1556] (node:1556) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.

Following this [tutorial](https://github.com/microsoft/vscode-docs/blob/master/docs/python/environments.md) I expected that the extension would notice when I create a virtual environment from the terminal but nothing happened. The new v env should also have been picked-up by the `Select Python Environment` command (shortcut on status bar), but were not listed among the choices .

marcdumais-work:
Seems `Python3` is not happy with the system's locale . If confirmed we should set this in image (likely done already in Ubuntu):

```
theia@56b525a48f4d:/home/project$ python3 -m flask run
Traceback (most recent call last):
  File "/usr/lib/python3.5/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/theia/.local/lib/python3.5/site-packages/flask/__main__.py", line 15, in <module>
    main(as_module=True)
  File "/home/theia/.local/lib/python3.5/site-packages/flask/cli.py", line 966, in main
    cli.main(prog_name="python -m flask" if as_module else None)
  File "/home/theia/.local/lib/python3.5/site-packages/flask/cli.py", line 586, in main
    return super(FlaskGroup, self).main(*args, **kwargs)
  File "/home/theia/.local/lib/python3.5/site-packages/click/core.py", line 696, in main
    _verify_python3_env()
  File "/home/theia/.local/lib/python3.5/site-packages/click/_unicodefun.py", line 124, in _verify_python3_env
    ' mitigation steps.' + extra
RuntimeError: Click will abort further execution because Python 3 was configured to use ASCII as encoding for the environment. Consult https://click.palletsprojects.com/en/7.x/python3/ for mitigation steps.

This system supports the C.UTF-8 locale which is recommended.
You might be able to resolve your issue by exporting the
following environment variables:

    export LC_ALL=C.UTF-8
    export LANG=C.UTF-8
```

Once I did Python's bidding, it rewarded me with a working app:

```
theia@56b525a48f4d:/home/project$ export LC_ALL=C.UTF-8
theia@56b525a48f4d:/home/project$     export LANG=C.UTF-8
theia@56b525a48f4d:/home/project$ 
theia@56b525a48f4d:/home/project$ python3 -m flask run
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
127.0.0.1 - - [12/Feb/2020 00:40:28] "GET / HTTP/1.1" 200 
```

theia-python-docker/Dockerfile
marcdumais-work: I do not have a strong opinion either way.

@jgroom33 As the original contributor of this image, do you have an opinion on whether it should have sudo available?

jgroom33: its nice to have sudo access to `apt install ...` but the goal of the image should probably be to have a more strict image creation. sudo should not be allowed
It should be the responsibility of the person doing the container packaging to modify this standard docker image and add the necessary packages for their use case.
Modifying an existing image that defaults to no sudo is easier than solving the problem of how to remove sudo and follow the standard.

marcdumais-work: @jgroom33 good points, thanks for the feedback. 

@vince-fugnitto let's remove sudo

vince-fugnitto: > @vince-fugnitto let's remove sudo

@marcdumais-work we should remove `sudo`, and do the following updates 
as well? https://github.com/theia-ide/theia-apps/pull/297#discussion_r377925621

marcdumais-work: > @marcdumais-work we should remove `sudo`, and do the following updates
> as well? [#297 (comment)](https://github.com/theia-ide/theia-apps/pull/297#discussion_r377925621)

:+1:

vince-fugnitto:
@marcdumais-work I updated the `baseImage` for the Dockerfile, removed the `gpg` keys, and also removed `sudo` access. I will need to find a way to add a newer version of Python since the base version is 3.5 and the [vscode-python](https://marketplace.visualstudio.com/items?itemName=ms-python.python) extension requires at least 3.6.

marcdumais-work:
> @marcdumais-work I updated the `baseImage` for the Dockerfile, removed the `gpg` keys, and also removed `sudo` access. I will need to find a way to add a newer version of Python since the base version is 3.5 and the [vscode-python](https://marketplace.visualstudio.com/items?itemName=ms-python.python) extension requires at least 3.6.


I hope there might be an better way, but last resort we could fetch the sources archive and build it as part of the image: https://medium.com/@rijaspk5195/install-python-3-6-4-on-debian-9-jessie-bb71cdc9ff35

vince-fugnitto:
> > @marcdumais-work I updated the `baseImage` for the Dockerfile, removed the `gpg` keys, and also removed `sudo` access. I will need to find a way to add a newer version of Python since the base version is 3.5 and the [vscode-python](https://marketplace.visualstudio.com/items?itemName=ms-python.python) extension requires at least 3.6.
> 
> I hope there might be an better way, but last resort we could fetch the sources archive and build it as part of the image: https://medium.com/@rijaspk5195/install-python-3-6-4-on-debian-9-jessie-bb71cdc9ff35

I updated the image to perform the steps you referenced (I could not find a simpler approach on debian). The image now has Python 3.6.4. (I tried updating to 3.7.0 but the build is too long).

theia-python-docker/Dockerfile
```diff
@@ -1,22 +1,43 @@
 ARG NODE_VERSION=10
-FROM node:${NODE_VERSION}-stretch
+FROM node:${NODE_VERSION}
 
+# Install Python 3 from source
 RUN apt-get update \
-    && apt-get install -y python python-dev python-pip \
-    && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/*
+    && apt-get upgrade -y \
```
marcdumais-work: Wow - I feel myself growing old as this builds and tests :)  Maybe there's at least a switch to skip the tests?

vince-fugnitto: > Wow - I feel myself growing old as this builds and tests :) Maybe there's at least a switch to skip the tests?

I saw the following comment from [stackoverflow](https://stackoverflow.com/questions/44708262/make-install-from-source-python-without-running-tests). I think there is a tradeoff between **build time** (skipping tests) and **better performance** in executing Python code.

At the moment I'm trying without the `--enable-optimizations` flag and it's building much more quickly.

vince-fugnitto: Maybe we can add a new build arg to control the behavior?

marcdumais-work: Looks the speed improvement to be expected when using the `--enable-optimizations` build switch is [between 1 and 20%](https://bugs.python.org/issue24915) depending on the scenario. This is of course nice - if we were publishing a generic python it might be worth it. But for this image here I'll take the shorter built-time tradeoff.

marcdumais-work:
> Following this [tutorial](https://github.com/Microsoft/vscode-docs/blob/master/docs/python/tutorial-flask.md) I expected that the extension would notice when I create a virtual environment from the terminal but nothing happened. The new v env should also have been picked-up by the `Select Python Environment` command (shortcut on status bar), but were not listed among the choices .

This problem is gone - now the venv is detected and proposed from the `select python environment` command

:+1:

theia-python-docker/Dockerfile
```diff
@@ -1,22 +1,43 @@
...
+RUN apt-get update \
+    && apt-get install -y python-dev python-pip \
+    && apt-get install -y python3-dev python3-pip \
+    && apt-get install -y yarn \
+    && apt-get clean \
+    && rm -rf /var/cache/apt/* \
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /tmp/*
 
+RUN adduser --disabled-password --gecos '' theia
```
marcdumais-work: If we do not have sudo I think we need to get rid of the Theia user and have root as default in the container. else it will not be possible to install anything. 

You'll need to explicitly create the `/home/theia` directory however, like it was probably done in the previous version of the image.

theia-python-docker/Dockerfile
```diff
@@ -1,22 +1,43 @@
...
+RUN chmod g+rw /home && \
+    mkdir -p /home/project && \
+    chown -R theia:theia /home/theia && \
```
marcdumais-work: remove along with the theia user

theia-python-docker/Dockerfile
```diff
@@ -1,22 +1,43 @@
...
+    chown -R theia:theia /home/project;
```
marcdumais-work: remove along with the theia user

vince-fugnitto:
I believe the image is ready for another round of review 😄

marcdumais-work:
I am not a python developer, so I am not sure how important this is. While testing using a tutorial, I installed something  through `pip` and got this warning:
```
You are using pip version 10.0.1, however version 20.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

I think the version of `pip` available in the DEBIAN repo is what it is, but we can potentially upgrade it during image compilation, using the suggested command.

vince-fugnitto:
> I think the version of `pip` available in the DEBIAN repo is what it is, but we can potentially upgrade it during image compilation, using the suggested command.

I now perform the update using:

```console
pip install --upgrade pip --user
```

vince-fugnitto:
> LGTM - nice work @vince-fugnitto

Thank you for the review!

PR #299, opened 2020-02-06
theia-electron: include default plugins

marechal-p:
Use a set of VS Code builtin extensions and some non-builtins to provide
more functionality to the Electron example IDE.

Signed-off-by: Paul Maréchal <paul.marechal@ericsson.com>


theia-electron/package.json
```diff
@@ -49,6 +52,8 @@
         "@theia/navigator": "next",
         "@theia/outline-view": "next",
         "@theia/output": "next",
+        "@theia/plugin-ext": "next",
```
akosyakov: we should clean up Theia extensions, i.e. all extensions replaced by built-in should go as well as typescript. @vince-fugnitto Could you help with it?

Actually package.json could be minimal like:
```json
{
  dependencies: {
     "@theia/plugin-ext-vscode": "next" ,
     "@theia/json": "next" ,
     "@theia/json": "preferences" 
  }
}
```
The rest will come transitively.

lmcbout:
In package.json, do we want to put in 
script :"start": "theia start --plugins=local-dir:./plugins",

Also in plugins section the dependencies of the builtins, add:
vscode-eslint

Also, why not using the electron-builder latest version, "electron-builder": "^22.0.1" (version 22.3.2),  it will create the AppImage like 
```
    Theia-0.3.17.AppImage
```

instead of (quote and space in the name)
```
  'Theia 0.3.17.AppImage'
```

marechal-p:
@spoenemann @akosyakov I only tested that plugins worked when bundled as an AppImage on Ubuntu. I would need help from people on other platforms to check if it works there too.

@lmcbout added the start script, even though the real value is in the bundles. I also fixed the name of the AppImage without updating electron-builder, I didn't want to fight with a CQ.

theia-electron/package.json
marechal-p: As much as I like small configs, when it comes to Theia extensions it is not really clear what to strip out and what to keep. Being explicit as the advantage of being very transparent.

lmcbout:
electron-buider has some fixes with latest version, we should aim to update it even it requires a CQ !
What dio you think

theia-electron/package.json
akosyakov: ok, but deprecated should be removed

vince-fugnitto:
> electron-buider has some fixes with latest version, we should aim to update it even it requires a CQ.
> What do you think?

There is no CQ required for this repo since it is not under the Eclipse Foundation.

Something like updating `electron-builder` can be done in a subsequent pull-request since it is an enhancement. I'd like to unblock this pull-request so we can continue with updating the necessary images and removing deprecated extensions.

marechal-p:
> There is no CQ required for this repo

Ok, updated `electron-builder`.

theia-electron/scripts/theia-electron-main.js
```diff
@@ -0,0 +1,7 @@
+const path = require('path')
```
marcdumais-work: @marechal-p 
I think we'll need to setup another plugins folder, using env var `THEIA_PLUGINS`, where the user can install additional ones, manually at first and through the new plugins widget soon. Any thoughts on that?

marcdumais-work:
Tested the `AppImage` on Ubuntu. @vince-fugnitto could you quickly validate that the plugins are loaded correctly on Windows and Mac?

marcdumais-work:
@marechal-p  While you're there, I think the `0.3.17` app version in `package.json` does not mean much - it was probably the Theia version at the time this app was created, but we do not want to step it every month. I suggest we change it to e.g. `0.0.1` or such. We should probably do this for all images.

(note: the Theia platform version is already reflected in the tag when we publish on `dockerhub`)

vince-fugnitto:
> Tested the `AppImage` on Ubuntu. @vince-fugnitto could you quickly validate that the plugins are loaded correctly on Windows and Mac?

I verified and it did not work correctly on **mac** or **win**.
After performing the updates https://github.com/theia-ide/theia-apps/commit/33c0027ccb6c9ba343d7fb5dc642e7afc91d3421, the electron build worked correctly and the plugins were successfully included/loaded.

theia-electron/package.json
```diff
@@ -72,7 +75,74 @@
...
+        "vscode-builtin-typescript-language-features": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/typescript-language-features-1.39.1-prel.vsix",
+        "vscode-builtin-vb": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/vb-1.39.1-prel.vsix",
+        "vscode-builtin-icon-theme-seti": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/vscode-theme-seti-1.39.1-prel.vsix",
+        "vscode-builtin-xml": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/xml-1.39.1-prel.vsix",
+        "vscode-builtin-yaml": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/yaml-1.39.1-prel.vsix",
+        "vscode-editorconfig": "https://github.com/theia-ide/editorconfig-vscode/releases/download/v0.14.4/EditorConfig-0.14.4.vsix",
+        "vscode-java-redhat": "https://github.com/redhat-developer/vscode-java/releases/download/v0.54.2/redhat.java-0.54.2.vsix",
```
vince-fugnitto: I don't think the following extensions are necessary in an example `JavaScript/TypeScript` desktop application.
I would add `eslint` however: https://github.com/eclipse-theia/theia/blob/master/package.json#L143

theia-electron/package.json
```diff
@@ -72,7 +75,74 @@
...
+        "vscode-builtin-json": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/json-1.39.1-prel.vsix",
+        "vscode-builtin-less": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/less-1.39.1-prel.vsix",
+        "vscode-builtin-log": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/log-1.39.1-prel.vsix",
+        "vscode-builtin-lua": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/lua-1.39.1-prel.vsix",
+        "vscode-builtin-make": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/make-1.39.1-prel.vsix",
+        "vscode-builtin-markdown": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/markdown-1.39.1-prel.vsix",
+        "vscode-builtin-markdown-language-features": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/markdown-language-features-1.39.1-prel.vsix",
```
vince-fugnitto: We should remove `vscode-builtin-markdown-language-features` in favor of `@theia/preview`.
The other images have been updated to do the same.

theia-electron/package.json
akosyakov: yes, it would be good just align it with VS Code, additional extensions should be then installed via UI

theia-electron/scripts/theia-electron-main.js
marechal-p: Done, added it as:

```ts
// Lookup inside the user's home folder for more plugins
process.env.THEIA_PLUGINS = `local-dir:${path.resolve(os.homedir(), '.theia', 'plugins')}`
```

marcdumais-work: For now uses will need to manually put the extra VS Code extensions they want in that folder. Mention in app's README?

akosyakov:
code-wise it looks good to me

theia-electron/scripts/theia-electron-main.js
vince-fugnitto: > For now users will need to manually put the extra VS Code extensions they want in that folder. Mention in app's README?

@marechal-p do you mind adding this small suggestion, then I think we can approve and merge 🍾

marechal-p: Done.

theia-electron/scripts/theia-electron-main.js
```diff
@@ -0,0 +1,13 @@
...
+const os = require('os')
+
+// Use a set of builtin plugins in our application.
+process.env.THEIA_DEFAULT_PLUGINS = `local-dir:${path.resolve(__dirname, '..', 'plugins')}`
+
+// Lookup inside the user's home folder for more plugins, and accept user-defined paths.
+process.env.THEIA_PLUGINS = [
+    process.env.THEIA_PLUGINS, `local-dir:${path.resolve(os.homedir(), '.theia', 'plugins')}`,
+].filter(Boolean).join(',')
```
marechal-p: @akosyakov IIRC when I tested this I was installing unpacked extensions there without any issues.

PR #323, opened 2020-03-26
rust: remove rust from CI

vince-fugnitto:
This commit removes `rust` from CI since it is somewhat fragile, and fails with the same error from the past: https://github.com/theia-ide/theia-apps/issues/253

---

@dwjbosman are you fine with the following approach? It looks like rust fails to build due to the same issue regarding the channel date. We can always add back CI for rust if/when the image becomes more stable.


vince-fugnitto:
@marcdumais-work do you think we can merge the pull-request?
I don't think the rust image is quite stable/maintained at the moment.

PR #334, opened 2020-04-06
Update README to mention theiaide/theia-full.

jchv:
New here, but I think this note may be useful for other newcomers trying to get a feel for Theia. (I suspect I may have some terminology mixed up here. Should I be calling the images Theia apps instead?)


jchv:
Ah, that looks nicer, closing.

vince-fugnitto:
> Ah, that looks nicer, closing.

You may always comment on the pull-request if you have ideas or suggestions 😃

Issue #332, opened 2020-04-05
docker run fails because percentage sign is not resolved


1kastner:
I tried to run theia as a docker image as it is described in the readme.
So it opened the PowerShell (which is I think the default command line for Windows) and copied line.
For some reason, the percentage sign was not resolved

```
PS C:\Users\XXX> docker run -it --init -p 3000:3000 -v "%cd%:/home/project:cached" theiaide/theia:next
Unable to find image 'theiaide/theia:next' locally
next: Pulling from theiaide/theia
aad63a933944: Pull complete
17551d40f9c7: Pull complete
1d4f35a66b6c: Pull complete
d4192b8fc2e1: Pull complete
4dda91c68cf9: Pull complete
af73feffc057: Pull complete
349e06e10445: Pull complete
3345427c5f33: Pull complete
Digest: sha256:6edac3eedf0e9ff221c36275170970cbe70eab10e9b55b6cb7c8cf9e474cf9ff
Status: Downloaded newer image for theiaide/theia:next
C:\Program Files\Docker\Docker\resources\bin\docker.exe: Error response from daemon: create %cd%: "%cd%" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'C:\Program Files\Docker\Docker\resources\bin\docker.exe run --help'.
```

Anyhow, with the command line (CMD) it was possible to start it.
Is it possible to add the information to the readme how to run the container both from PS and CMD?

vince-fugnitto:
@1kastner thank you for the issue! 
I unfortunately do not have a Windows machine in order to verify the behavior for PowerShell, so if you can do a quick fix it'd be really appreciated :)

@marechal-p perhaps you could do a quick test?

PR #338, opened 2020-04-08
Point out that the commands are for the CMD

1kastner:
These commands fail in the PowerShell, see #332 for reference

PR #339, opened 2020-04-08
Add reference to PowerShell

1kastner:
PowerShell can also understand the dollar-and-bracket expression, so no need to send them to the CMD. See https://stackoverflow.com/questions/41485217/mount-current-directory-as-a-volume-in-docker-on-windows-10/41489151#41489151 for reference. This pull request is based on #332.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,19 +28,19 @@ At the moment Theia is still in [active development](https://github.com/theia-id
 
 This script pulls the image and runs Theia IDE on http://localhost:3000 with the current directory as a workspace. The option of `--init` is added to fix the [defunct process problem](https://github.com/theia-ide/theia-apps/issues/195).
 
-    # Linux or macOS
+    # Linux, macOS, or PowerShell
     docker run -it --init -p 3000:3000 -v "$(pwd):/home/project:cached" theiaide/theia:next
     
-    # Windows
+    # Windows (cmd.exe)
     docker run -it --init -p 3000:3000 -v "%cd%:/home/project:cached" theiaide/theia:next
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,7 +28,7 @@ At the moment Theia is still in [active development](https://github.com/theia-id
 
 This script pulls the image and runs Theia IDE on http://localhost:3000 with the current directory as a workspace. The option of `--init` is added to fix the [defunct process problem](https://github.com/theia-ide/theia-apps/issues/195).
 
-    # Linux or macOS
+    # Linux, macOS, or PowerShell
```
vince-fugnitto, line 31: I'm not sure how we should organize the following information, but I think we should keep the categories based on operating-system and not display PowerShell at this level to avoid confusion.

I think we should create a single pull-request with the following changes and those present in [#388](https://github.com/theia-ide/theia-apps/pull/338/files), keep the original categories, and under Windows we display the two options, one for cmd and one for powershell.

marechal-p: PowerShell can be installed on Ubuntu as well... But it is more commonly used on Windows indeed.

marechal-p: I think it is fine as it is right now, at least it doesn't bother me.

1kastner: I mentioned PowerShell separately for the reason @marechal-p mentioned - the open source version can be installed on other operating systems as well. If you ask me personally, the line referring to CMD could rather be dropped because Microsoft pushes people to use PowerShell more and more.

marechal-p:
Hm, could you please squash your commits though? Will be ready to merge then.

1kastner:
Sure!

PR #330, opened 2020-04-03
add vsx-registry to ci builds

emdfonseca:
#### What it does
Add @theia/vsx-registry dependency to latest and next package.json files of every app that is part of the CI pipeline (theia-cpp-docker theia-dart-docker theia-docker theia-full-docker theia-go-docker theia-java-docker theia-python-docker theia-rust-docker theia-swift-docker).

Closes #325 

#### How to test
Trigger a build and let it finish. Everything builds without any issue.

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/eclipse-theia/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to review in accordance with [the review guidelines](https://github.com/eclipse-theia/theia/blob/master/doc/pull-requests.md#reviewing)

Signed-off-by: Emanuel Fonseca <emdfonseca@gmail.com>

PR #343, opened 2020-04-13
[CI] [Rust] Add-back the Rust image to CI

marcdumais-work:
We've been having issues for a while with the Rust image in CI. We're temporarily removed it so that CI can pass, but would like to being it back when the issues are fixed. 

For a while, it looked like the Rust tools were causing the CI issue. A recent update seems to have remove this problem: https://github.com/theia-ide/theia-apps/pull/330#issuecomment-611981825 . So I thought I would give it a try and see what other issues there might be. 

In addition to adding the image back into CI, this PR:
- Removes deprecated @theia/editorconfig (prevented app from starting, though it compiled)
- Adds environment variable so that plugins (VS Code extensions) defined  in package.json are detected/used

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,7 @@ env:
   - NPM_TAG=next IMAGE_NAME=theia-swift NODE_VERSION=10
   - NPM_TAG=latest IMAGE_NAME=theia-cpp NODE_VERSION=10.15.3
   - NPM_TAG=next IMAGE_NAME=theia-cpp NODE_VERSION=10.15.3
-  # skip the `rust` CI build due to known failures: https://github.com/theia-ide/theia-apps/issues/253
-  # - NPM_TAG=next IMAGE_NAME=theia-rust NODE_VERSION=10.15.3
+  - NPM_TAG=next IMAGE_NAME=theia-rust NODE_VERSION=10.15.3
   - NPM_TAG=latest IMAGE_NAME=theia-dart NODE_VERSION=10
   - NPM_TAG=next IMAGE_NAME=theia-dart NODE_VERSION=10
   - NPM_TAG=latest IMAGE_NAME=theia-https NODE_VERSION=10 ENV_VARS="-e token=" PORT=10443
--- a/theia-rust-docker/Dockerfile
+++ b/theia-rust-docker/Dockerfile
@@ -23,8 +23,10 @@ ADD next.package.json $THEIA_RUST_APP_PATH/package.json
 RUN git clone https://github.com/eclipse-theia/theia-cpp-extensions.git /root/theia-cpp-extension
 RUN cp -R /root/theia-cpp-extension/packages $THEIA_RUST_APP_PATH/.
 RUN cp -Rv /root/theia-cpp-extension/dev-packages/* $THEIA_RUST_APP_PATH/packages/.
-RUN . /root/nvm_setup.sh && cd $THEIA_RUST_APP_PATH && yarn --cache-folder $THEIA_RUST_APP_PATH/ycache 
-RUN . /root/nvm_setup.sh && cd $THEIA_RUST_APP_PATH && yarn theia build
+RUN . /root/nvm_setup.sh && cd $THEIA_RUST_APP_PATH && yarn --cache-folder $THEIA_RUST_APP_PATH/ycache && \
+    . /root/nvm_setup.sh && cd $THEIA_RUST_APP_PATH && yarn theia build && \
+    . /root/nvm_setup.sh && cd $THEIA_RUST_APP_PATH && yarn theia download:plugins && \
+    rm -rf $THEIA_RUST_APP_PATH/ycache
 
 # when creating a container a volume is mapped to /root/rust
 # to allow to make persistent changes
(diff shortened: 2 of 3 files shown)


emdfonseca:
My only suggestion would be to change the rust channel to "nightly". Pinning to a specific date will fail as soon as the version becomes unavailable.

theia-rust-docker/Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN cp -R /root/theia-cpp-extension/packages $THEIA_RUST_APP_PATH/.
 RUN cp -Rv /root/theia-cpp-extension/dev-packages/* $THEIA_RUST_APP_PATH/packages/.
 RUN . /root/nvm_setup.sh && cd $THEIA_RUST_APP_PATH && yarn --cache-folder $THEIA_RUST_APP_PATH/ycache 
 RUN . /root/nvm_setup.sh && cd $THEIA_RUST_APP_PATH && yarn theia build
+RUN . /root/nvm_setup.sh && cd $THEIA_RUST_APP_PATH && yarn theia download:plugins
```
emdfonseca, line 28: We could merge all RUN commands into a single RUN to make the image smaller.

.travis.yml
```diff
@@ -32,8 +32,7 @@ env:
   - NPM_TAG=next IMAGE_NAME=theia-swift NODE_VERSION=10
   - NPM_TAG=latest IMAGE_NAME=theia-cpp NODE_VERSION=10.15.3
   - NPM_TAG=next IMAGE_NAME=theia-cpp NODE_VERSION=10.15.3
-  # skip the `rust` CI build due to known failures: https://github.com/theia-ide/theia-apps/issues/253
-  # - NPM_TAG=next IMAGE_NAME=theia-rust NODE_VERSION=10.15.3
+  - NPM_TAG=next IMAGE_NAME=theia-rust NODE_VERSION=10.15.3
```
emdfonseca, line 35: Is there any reason not to have the latest tag as well for rust?

marcdumais-work:
> My only suggestion would be to change the rust channel to "nightly". Pinning to a specific date will fail as soon as the version becomes unavailable.

Thanks for the feedback. Do you know if we can expect that there could eventually be breaking changes if we use "nightly". e.g. it going to the next major version, breaking some backward compatibility and necessitating that we update the image?

emdfonseca:
I'm not developing on rust so I don't have an answer. Anyway, the stable channel is not building. We can only choose between pinned or unpinned nightly. I would go for the unpinned nightly and evaluate over time whether it's stable enough.

My main argumentation is:
1. CI only builds if there are changes in the folder. I guess the rust image will not change that often and if it does there's a chance to be done by a rust developer who might quickly identity the problem and evaluate other/better solutions
2. By pinning to a date we know it will fail as soon as the build is unpublished (100% predictable). By leaving without a date there's the chance it will never fail :) (highly optimistic still the chances are slightly higher)

Any rust developer amongst us that would like to share his/her thoughts?

marcdumais-work:
cc: @dwjbosman Maybe you have some feedback about the rust channel discussion?

marcdumais-work:
Thanks @emdfonseca - I am also not a rust developer but your arguments make sense to me. I have pinged the original author of the rust image - I hope he might have some valuable feedback as well.

> I'm not developing on rust so I don't have an answer. Anyway, the stable channel is not building. We can only choose between pinned or unpinned nightly. I would go for the unpinned nightly and evaluate over time whether it's stable enough.
> 
> My main argumentation is:
> 
> 1. CI only builds if there are changes in the folder. I guess the rust image will not change that often and if it does there's a chance to be done by a rust developer who might quickly identity the problem and evaluate other/better solutions

This is generally true but I think with one exception: the nightly build. It permits having a version of the images, built against the most recent eclipse-theia "master" branch. I do not think this invalidates your arguments however.

> 2. By pinning to a date we know it will fail as soon as the build is unpublished (100% predictable). By leaving without a date there's the chance it will never fail :) (highly optimistic still the chances are slightly higher)

I am willing to give this a try :+1:

.travis.yml
marcdumais-work: Good question. IIRC when this image was introduced, it relied on some "just-released" Theia dependencies, not yet available as `latest`. I think it was `@theia/cpp`. So we could introduce a `latest` version of this image when its CI is "stabilised".

theia-rust-docker/Dockerfile
marcdumais-work: done

marcdumais-work:
CI has passed with the latest PR version.

@emdfonseca so far so good about using "nightly" channel - let's keep it like that and see how long it works. I have also addressed your other comments.

@vince-fugnitto you want to have a look? There are not much changes.

marcdumais-work:
Thanks for the help @emdfonseca. 

I noticed after the fact  that you got started by asking a question on our Spectrum forum, and then you came here to submit a PR related to your question, rather then waiting for someone else to take care of the task. I think that's awesome. 

I wanted to ask if you have a general interest in Theia applications or if this was a little detour you needed to make in order to be able to fulfil l a use-case of yours? If you have a general interest, I can add you as contributor to the repo and you'd be welcome to help if you like. Please let us know.

emdfonseca:
Hi @marcdumais-work, it was mostly a consequence of my continuous journey of learning and experimenting. :) I might come back from time to time to support the project but nothing planned so far. If I end up supporting more I will come back to you. Thanks!

marcdumais-work:
> Hi @marcdumais-work, it was mostly a consequence of my continuous journey of learning and experimenting. :) I might come back from time to time to support the project but nothing planned so far. If I end up supporting more I will come back to you. Thanks!

Understood, thanks @emdfonseca

PR #348, opened 2020-04-17
[next][cpp] switch to vscode-clangd language client

marcdumais-work:
We can only do this change for `next` at the moment because some recently merged features are required to run `vscode-clangd` in Theia applications. We can do `latest` after this month's release.

- remove `@theia/cpp`
- add `vscode-clangd` 
- README update

image

image

### DIFF
--- a/theia-cpp-docker/README.md
+++ b/theia-cpp-docker/README.md
@@ -1,16 +1,19 @@
-## Theia cpp Docker
+# Theia cpp Docker
 
 A containerized Theia-based C/C++ demo IDE, including commonly used tools:
+
 - latest clangd Language Server (nightly build)
 - latest stand-alone clang-tidy static analyser (nightly build)
 - GDB 8.1 (from Ubuntu repo)
-- cmake 3.10.2 (from Ubuntu repo)
+- cmake 3.17.0
 
 The included Theia-based IDE application has the following notable features
-- [@theia/cpp] Language-server built-in clang-tidy static analyser integration. Will analyse files opened in the IDE's editors and report problems for configured rules. See [README](https://github.com/theia-ide/theia/tree/master/packages/cpp#using-the-clang-tidy-linter) for more details, including related preferences
+
+- [deprecated][latest][@theia/cpp] Language-server built-in clang-tidy static analyser integration. Will analyse files opened in the IDE's editors and report problems for configured rules. See [README](https://github.com/theia-ide/theia/tree/master/packages/cpp#using-the-clang-tidy-linter) for more details, including related preferences. Note: this LSP client is _deprecated_ and will be replaced once `vscode-clangd` works well with _latest_ Theia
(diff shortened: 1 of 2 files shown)


marcdumais-work:
One slightly weird thing: I am pretty sure we used to need a special docker command-line argument so that debugging with GDB would work in Docker. It's apparently no longer required - it would be good to independently confirm.

theia-cpp-docker/README.md
```diff
@@ -1,16 +1,19 @@
-## Theia cpp Docker
+# Theia cpp Docker
```
vince-fugnitto, line 1: I was thinking, should we update the readme title to **Theia C/C++ Docker** so it's clearer?

theia-cpp-docker/next.package.json
```diff
@@ -126,6 +123,7 @@
         "vscode-builtin-icon-theme-seti": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/vscode-theme-seti-1.39.1-prel.vsix",
         "vscode-builtin-xml": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/xml-1.39.1-prel.vsix",
         "vscode-builtin-yaml": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/yaml-1.39.1-prel.vsix",
-        "vscode-editorconfig": "https://github.com/theia-ide/editorconfig-vscode/releases/download/v0.14.4/EditorConfig-0.14.4.vsix"
+        "vscode-editorconfig": "https://github.com/theia-ide/editorconfig-vscode/releases/download/v0.14.4/EditorConfig-0.14.4.vsix",
+        "vscode-clangd": "https://open-vsx.org/api/llvm-vs-code-extensions/vscode-clangd/0.0.21/file/llvm-vs-code-extensions.vscode-clangd-0.0.21.vsix"
```
vince-fugnitto, line 127: In general, I first grouped builtins alphabetically, and then grouped extensions alphabetically.
By that same design I think it should come before `vscode-editorconfig`.

marcdumais-work:
> I'm fine with updating the changes, should we merge before the release or wait till after to update both images? I'm not sure it's pressing at the moment?

It would be less work the way you suggest but I'd like to have an example using `vscode-clangd` ASAP.

theia-cpp-docker/next.package.json
marcdumais-work: done

theia-cpp-docker/README.md
marcdumais-work: Just that entry or the image name as well? The next line mentions "C/C++" so I think it's not too confusing?

vince-fugnitto: I was thinking just the title but I'm fine either way (and it's not related to your changes).

marcdumais-work:
Thanks for the review @vince-fugnitto. Merging.

marcdumais-work:
> Thanks for the review @vince-fugnitto. Merging (once CI passes).

Issue #368, opened 2020-06-05
[Feature Request] Add dart/flutter support to full image


chrbayer:
In the full image support for the dart language is missing. Please add it to the full image as well, so that is really containing support for all available languages. If support for flutter is available, it should go to the full image, too :-)

chrbayer:
Dart SDK is now included in the full image, for me this issue can be closed.

PR #369, opened 2020-06-05
Add dart sdk to full image and install dart and flutter extensions to next image

chrbayer:
Please review. One drawback is that the dart SDK is installed to Dockerimage for stable and next image.

### DIFF
--- a/theia-full-docker/next.package.json
+++ b/theia-full-docker/next.package.json
@@ -88,6 +88,7 @@
         "vscode-builtin-lua": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/lua-1.39.1-prel.vsix",
         "vscode-builtin-make": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/make-1.39.1-prel.vsix",
         "vscode-builtin-markdown": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/markdown-1.39.1-prel.vsix",
+        "vscode-builtin-markdown-language-features": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/markdown-language-features-1.39.1-prel.vsix",
         "vscode-builtin-merge-conflicts": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/merge-conflict-1.39.1-prel.vsix",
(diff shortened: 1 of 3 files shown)


chrbayer:
This fixes #368.

theia-full-docker/next.package.json
```diff
@@ -56,7 +56,8 @@
         "**/vscode-json-languageserver/**/vscode-languageserver": "6.0.0-next.1"
     },
     "devDependencies": {
-        "@theia/cli": "next"
+        "@theia/cli": "next",
+        "@theia/debug": "next"
```
vince-fugnitto, line 60: @chrbayer can you explain why `@theia/debug` was added as a `devDependency`? (it is already added as a dep)

chrbayer: I have not noticed that "@theia/debug" is already in included in the dependencies, just copied it from the dart image. It will be removed :-)

vince-fugnitto:
> I have not added it to latest because it did not want to make this change for latest, the idea was to wait until next is getting latest. Do you think I should make the change in latest, too?

@chrbayer  don’t see why we wouldn’t add the functionality to both images, they both should have the proper functionality to make use of the new vscode extensions introduced by the pull-request.

Please be sure to properly squash your commits when you’re ready for a re-review :)

chrbayer:
Thanks for the review :-) I have added the plugins to the latest image, too. The pull request should be complete now.

But there is a problem with downloading repgrep, I have the same problem with ci and my local builds... There is a new repgrep version released, but I can't see why this breaks downloading the previous version...

chrbayer:
This should be final now if nothing is found in the review.

marcdumais-work:
> But there was a problem with downloading repgrep, I had the same problem with ci and my local builds... This seem to work again.

Indeed - the latest `ripgrep` was momentarily broken. It's now fixed. See https://github.com/microsoft/vscode-ripgrep/issues/13

marcdumais-work:
Thanks @chrbayer . @vince-fugnitto ready for a final review on this one?

chrbayer:
@marcdumais-work Ready from my side. Thanks in advance!

vince-fugnitto:
@chrbayer do you mind squashing your commits (into a single commit) so we can accept the changes? :)

chrbayer:
@vince-fugnitto Of course, I will do it :-)

marcdumais-work:
@chrbayer I think the commit message needs an update to reflect the current content. I suggest something like: 
`[full] Add dart to latest and next images` 
or simply:
`[full] Add dart`

marcdumais-work:
Awesome, thanks for the update @chrbayer 

@vince-fugnitto Can you have another look, and if no further comments, merge?

PR #370, opened 2020-06-06
Added Travis CI jobs to publish installers

thegecko:
Signed-off-by: thegecko <rob.moran@arm.com>

This PR introduces electron builds and publishing from TravisCI for the master branch.

I've already created a test publish of the installers here: https://github.com/theia-ide/theia-apps/releases/tag/v0.1.0

__Note:__ These installer binaries are untested and currently unsigned. I will create a separate PR later to introduce signing for macOS and Windows.

## Changes

- Added Travis CI jobs to build installers for macOS/Windows/Linux in addition to existing docker jobs which only run on master
- Updated `theia-electron` project to use `latest` theia by default to be more stable for users
- Updated git repository URLs in `theia-electron` project to point at this repo (as it's used by electron-builder for publishing)
- Automatic publishing of installers to GitHub when a draft release exists matching the current version
- Addition of an [entry to the wiki](https://github.com/theia-ide/theia-apps/wiki/Publishing-a-release) which outlines how to publish a release

## Decisions to confirm
- theia tag to use for a release (currently this uses what is set in the `package.json` file which is `latest` and __not__ `next` nor an explicit version)
- publishing is done on GitHub (should it be elsewhere?)
- publishing is to the `theia-apps` repo and not `eclipse-theia`
- do we need to consider different release channels in future?

__Notes:__ 

- Merging this PR should automatically add binaries to a release that has been drafted for `v1.2.0`.
- There is a build error for [theia-next builds](https://travis-ci.org/github/theia-ide/theia-apps/builds/695292595) on master which I believe is unrelated to these changes

cc @akosyakov @marcdumais-work

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,15 @@ branches:
   only:
   - master
 
-language: generic
+language: node_js
+node_js: "10"
 
 addons:
   apt:
     update: true
     sources:
     - ubuntu-toolchain-r-test
+    packages: libxkbfile-dev
     chrome: stable
 
 env:
@@ -40,10 +42,21 @@ env:
   - NPM_TAG=latest IMAGE_NAME=theia-php NODE_VERSION=10
   - NPM_TAG=next IMAGE_NAME=theia-php NODE_VERSION=10
 
+jobs:
+  # add installer builds to jobs matrix
+  include:
+    - &publish
+      env:
+      install: skip
+      script: cd theia-electron && yarn && yarn package
+      deploy: skip
+    - <<: *publish
+      os: osx
+    - <<: *publish
+      os: windows
+
 before_install:
-- nvm install 10
-- nvm use 10
-- npm install -g eclint
+- npm install -g yarn eclint
 - eclint check "**/*" || echo "Warning! Formatting errors encountered..."
 
 install:
--- a/build_container.sh
+++ b/build_container.sh
@@ -24,7 +24,7 @@ cd "$IMAGE_NAME-docker"
 IMAGE="theiaide/$IMAGE_NAME"
 IMAGE_TAG="$IMAGE":$(npm view "@theia/core@$NPM_TAG" version)
 echo $IMAGE_TAG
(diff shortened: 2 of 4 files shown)


thegecko:
Confirmed the build error is due to https://github.com/eclipse-theia/theia/issues/7974

akosyakov:
I've tried pre-release for Mac and it did not work for me:
Screenshot 2020-06-08 at 08 57 51
Screenshot 2020-06-08 at 08 58 28

kittaakos:
> I've tried pre-release for Mac and it did not work for me:

The app isn't signed; it's expected, unfortunately.

akosyakov:
> The app isn't signed; it's expected, unfortunately.

The second error is not about singing. It does not start to me if I even override it.

kittaakos:
> The second error is not about singing.

I know; we have this error in the electron app since we ship with the VS Code extension support. It happens every time when the app start has failed. It is this: https://github.com/eclipse-theia/theia/issues/7113

kittaakos:
> It does not start to me if I even override it.

I do not understand this part, can you explain what do you many by _even override it_?

thegecko:
Travis CI job is now passing, this should be ready for review.

marcdumais-work:
> The second error is not about singing. It does not start to me if I even override it. 

My interpretation: there is a way to bypass the check for signing, but the packaged app still does not work correctly

marcdumais-work:
I think we'll need to update the README to really spell-out what this application is - an example desktop Theia application. I'll suggest some updates to the package.json as well for the same reason.

.travis.yml
```diff
@@ -40,10 +42,21 @@ env:
...
+jobs:
+  # add installer builds to jobs matrix
+  include:
+    - &publish
+      env:
+      install: skip
+      script: cd theia-electron && yarn && yarn package
+      deploy: skip
+    - <<: *publish
+      os: osx
```
marcdumais-work, line 54: @thegecko It looks unlikely we'll want to publish from here for Windows/Mac, since the signing will not work here. I am happy to keep these entries here for now and remove them when we have this working elsewhere. Sounds like a plan?

marcdumais-work:
@thegecko  One little thing before merging, if possible: it would be nice to squash the 3 commits into a single one.

PR #372, opened 2020-06-13
[cpp,go,python,full] Sync single language images with full image if possible and update versions

chrbayer:
Where possible sync the single language images and the full images and the update versions of development tools in full and other images when in sync:

- cpp: Update cmake to 3.17.4, add tools for 32 bit development
- go: Update to golang 1.14.4
- java: Update to openjdk 11
- python: Update to latest 3.8 release (3.8.3)
- php: Update to latest 7.4 release (7.4.7)


marcdumais-work:
Thanks for this PR, @chrbayer - This looks like a nice update!

theia-cpp-docker/Dockerfile
```diff
@@ -101,16 +101,23 @@ RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
...
+                       gcc-multilib \
+                       g++-multilib \
                        gdb && \
-    ln -s /usr/bin/clangd-$LLVM /usr/bin/clangd && \
+    ln -s /usr/bin/clang-$LLVM /usr/bin/clang && \
+    ln -s /usr/bin/clang++-$LLVM /usr/bin/clang++ && \
+    ln -s /usr/bin/clang-cl-$LLVM /usr/bin/clang-cl && \
+    ln -s /usr/bin/clang-cpp-$LLVM /usr/bin/clang-cpp && \
     ln -s /usr/bin/clang-tidy-$LLVM /usr/bin/clang-tidy && \
-    rm -rf /var/lib/apt/lists/*
```
marcdumais-work, line 107: Why remove this "cleanup" line?

theia-cpp-docker/Dockerfile
```diff
@@ -101,16 +101,23 @@ RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
...
+    ln -s /usr/bin/clangd-$LLVM /usr/bin/clangd
 
 # Install latest stable CMake
-RUN wget "https://cmake.org/files/v3.17/cmake-3.17.0-Linux-x86_64.sh" && \
-    chmod a+x cmake-3.17.0-Linux-x86_64.sh && \
-    ./cmake-3.17.0-Linux-x86_64.sh --prefix=/usr/ --skip-license && \
-    rm cmake-3.17.0-Linux-x86_64.sh
+ARG CMAKE_VERSION=3.17.3
```
marcdumais-work, line 115: Move at the top of the file, along with other similar version-related docker ARGs?

theia-full-docker/next.package.json
```diff
@@ -124,6 +129,7 @@
         "vscode-builtin-icon-theme-seti": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/vscode-theme-seti-1.39.1-prel.vsix",
         "vscode-builtin-xml": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/xml-1.39.1-prel.vsix",
         "vscode-builtin-yaml": "https://github.com/theia-ide/vscode-builtin-extensions/releases/download/v1.39.1-prel/yaml-1.39.1-prel.vsix",
+        "vscode-clangd": "https://open-vsx.org/api/llvm-vs-code-extensions/vscode-clangd/0.0.21/file/llvm-vs-code-extensions.vscode-clangd-0.0.21.vsix",
```
marcdumais-work, line 132: Good idea to add `vscode-clangd`. You should remove `@theia/cpp` from the `dependencies` section, since they both do similar. 

Can you do the same change in the `latest.package.json` and the `theia-cpp` image?

theia-cpp-docker/Dockerfile
chrbayer: I removed the line at this position because I thought it would be a good idea to do the cleanup only once after the last install with apt:

```
# Theia application
##Needed for node-gyp, nsfw build
RUN apt-get update && apt-get install -y python build-essential && \
  apt-get clean && \
  apt-get autoremove -y && \
  rm -rf /var/cache/apt/* && \
  rm -rf /var/lib/apt/lists/* && \
  rm -rf /tmp/*
```

theia-cpp-docker/Dockerfile
chrbayer: This is some kind of design decision... I just kept the style with I found in the file, there where all version defines right before the code block which installs them. This way it is easy to comment complete blocks out to personalize the image to ones need...
But if you prefer, I can put all of them at the top of the file, this way it is easier to see what needs to be updated :-)

theia-full-docker/next.package.json
chrbayer: I'm not so familiar with the different plugins/extensions from different sources... For me it was hard to judge what is really needed and what not... What plugin is responsible for the colorization of the source code? I thought, this is not the clangd extension... I can give it a try, if I see no problems I can remove ```@theia/cpp```.

chrbayer: @marcdumais-work I have now tested the image without `@theia/cpp` and everything works as expected, even clanmg-tidy. So I updated the the clangd extension to the latest available version in the repository and made the chances as you proposed.

theia-cpp-docker/Dockerfile
marcdumais-work: This way of doing it once at the end is good for the final tidiness of the image's file system, but IIUC has a drawback: unless unwanted files are removed from the image in the same docker command (i.e. same image layer) they were created, they will still be in the image but hidden. 

So ideally such cleanups would happen for every `apt-get` command. However I can't say that we've been very careful so far to do this everywhere in our images, so it's not necessarily a deal breaker for this PR.  

WDYT?

theia-cpp-docker/Dockerfile
marcdumais-work: Fair enough!

theia-full-docker/next.package.json
marcdumais-work: Good thinking - indeed it used to be that `@theia/cpp` was providing the Textmate grammar responsible for syntax highlighting for `C/C++`. `vscode-clangd` does not provide it, since it's assumed to be present by default in VS Code. 

The grammar  is now contributed by this "built-in" VS Code extension, that we added a little while ago: https://github.com/theia-ide/theia-apps/pull/372/files#diff-fec21bf02d71557187accd444db76d3cL67

theia-cpp-docker/Dockerfile
chrbayer: This is a good point which have not thought about. It will have only a minimum effect on build times, so I'm going to copy the cleanup-block after each apt-get usage.

theia-full-docker/Dockerfile
```diff
@@ -220,40 +321,60 @@ ENV PATH $DART_SDK/bin:/theia/.pub-cache/bin:$PATH
...
+    GOPATH=/home/project
 
-ENV GOPATH=/home/project
-ENV PATH=$PATH:$GOPATH/bin
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
-ENTRYPOINT [ "yarn", "theia", "start", "/home/project", "--hostname=0.0.0.0" ]
+ENTRYPOINT [ "node", "/home/theia/src-gen/backend/main.js", "/home/project", "--hostname=0.0.0.0" ]
+# ENTRYPOINT [ "yarn", "theia", "start", "/home/project", "--hostname=0.0.0.0" ]
```
marcdumais-work, line 380: Remove commented-out line?

marcdumais-work:
Thanks @chrbayer - I just did another pass and found just one minor thing. I need to test a bit, but so far the changes look good! I'll let you know how it goes with the tests,

theia-full-docker/Dockerfile
marcdumais-work: BTW, I am curious: did you find a functional difference between the two forms of `ENTRYPOINT` above?

theia-cpp-docker/Dockerfile
marcdumais-work: > Changes made and pushed, the size of the final image stays the same.

That part is a bit unexpected. I wonder why? Any theories? I wonder if you were to remove all these cleanups related to installing Debian packages, including the final one, would the image size increase?

theia-full-docker/Dockerfile
chrbayer: I'm going to delete the commented line after you review and tests are finished, maybe there will be some more changes. I didn't noticed any difference between the two lines, just took the change out of the cpp docker script.

chrbayer:
@marcdumais-work Thank you very much for reviewing and testing. Please tell me if all tests and reviews are finished, so that I can include all findings in the PR. Should I make a sperate commit or just re-push fixed?

marcdumais-work:
> Should I make a sperate commit or just re-push fixed?

It's fine to force-push your fix(es). 

Normally we prefer squashing the commits of PRs but in this case, looking a the commit messages, the breakdown looks logical and if you think you've maintained the changes pretty well insulated, each in their own commit, I have no problem retaining the various commits you have in the PR currently. 

Though if we conclude it will make no difference in image size, we could return to a single cleanup at the end for apt temp files (i.e. drop last commit)

marcdumais-work:
This PR introduces lots of changes, so it will be hard to test everything. I've done some sanity testing, just to confirm that the main things seem to still work. No issue found!

[full]: 
- [x] C++ debug (hit breakpoint, inspect value of local variables). Used this [example workspace](https://github.com/eclipse-theia/theia-cpp-extensions/tree/master/examples/cpp-debug-workspace)
- [x] C++ Language server working (completions, Hover info, problem markers created for syntax issues, cleared when fixed)
- [x] Python debug and language server working. Used this [tutorial](https://code.visualstudio.com/docs/python/python-tutorial) as base.
- [x] Go debug and language server working

[cpp]
- [x] C++ debug (hit breakpoint, inspect value of local variables). Used this [example workspace](https://github.com/eclipse-theia/theia-cpp-extensions/tree/master/examples/cpp-debug-workspace)
- [x] C++ Language server working (completions, Hover info, problem markers created for syntax issues, cleared when fixed)

[python]
- [x] Python debug and language server working. Used this [tutorial](https://code.visualstudio.com/docs/python/python-tutorial) as base.
- [x] Go debug and language server work

marcdumais-work:
@chrbayer Ready for your final update. 

Can you let me know if anything substantial changes vs the current version, that would make it worth it to redo some of the testing?

chrbayer:
@marcdumais-work Thank you very much for all you support! The only changes I did are the two proposed changes:
1. Remove the commented line at the end of the cpp and full docker files
2. Removing the apt cleanup after each step and do the cleanup only at the end of the script

chrbayer:
For me this PR should now be complete :-)

PR #373, opened 2020-06-14
[full] Add Swift SDK to full image

chrbayer:
Add swift language tools to full image. If you want to use the swift REPL, these options have to be provided to docker: --cap-add=SYS_PTRACE --security-opt seccomp=unconfined

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Many applications in the repository now include support for running VS Code exte
 
 - #### Build Options:
     - `--init` injects an instance of [tini](https://github.com/krallin/tini) in the container, that will wait-for and reap terminated processes, to avoid leaking PIDs.
-    - `--security-opt seccomp=unconfined` enables running without [the default seccomp profile](https://docs.docker.com/engine/security/seccomp/) for debugging.
+    - `--security-opt seccomp=unconfined` enables running without [the default seccomp profile](https://docs.docker.com/engine/security/seccomp/) for debugging. This option is also required if the swift REPL is needed.
 
 ## Contributing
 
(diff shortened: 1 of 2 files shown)


marcdumais-work:
> If you want to use the swift REPL, these options have to be provided to docker: --cap-add=SYS_PTRACE --security-opt seccomp=unconfined

We used to need the same or similar docker options, so that we could debug C/C++ programs in a container using GDB. However we recently noticed that it now works by default, without any special option. IIRC it was due to a docker change, re-classifying some options as being safe, and so they are now enabled by default. 

Docker version I use: `Docker version 19.03.2, build 6a30dfc`

@chrbayer assuming you have a relatively recent docker version, could you confirm if the mentioned CLI options are still required for use of the swift REPL? If so a mention in the README would be good.

chrbayer:
I'm using ```Docker version 19.03.11-ce, build 42e35e61f3``` and I'm absolutely sure this options are still necessary for this version because it took me some time to google them and he REPL was crashing all the time. I will include them in the README.

marcdumais-work:
Looks like the `full latest` image's Theia app was not responding during tests. I've restarted CI for that one only - I see no reason why this PR would cause this issue - probably just a glitch.

chrbayer:
In that last time, I had often problems with the CI and had to start it all over until all images were built correctly...

marcdumais-work:
> In that last time, I had often problems with the CI and had to start it all over until all images were built correctly...

Unfortunately there's a know issue for contributors who submit PRs from their own fork - the Travis secure environment variables are not used then, for security reasons, and you'll often be unable to complete the build. The symptom usually an error related to `vscode-ripgrep`. 

You've had a few contributions already - maybe I can give write access to the repo and then you can push your PR branch directly on this repo before creating a PR - that would avoid the issue. 

If this sounds good to you, please ping me before opening your next PR and we can give it a shot.

marcdumais-work:
> Looks like the `full latest` image's Theia app was not responding during tests. I've restarted CI for that one only - I see no reason why this PR would cause this issue - probably just a glitch.

CI passed this time :+1:

PR #403, opened 2020-09-09
python: update python image

vince-fugnitto:
**Description**

The following commit updates the `theia-python-docker` image to a multi-stage build approach, where we build the application in the first stage and copy it over to the second (first layer is not preserved). 

### DIFF
--- a/theia-python-docker/Dockerfile
+++ b/theia-python-docker/Dockerfile
@@ -1,4 +1,21 @@
 ARG NODE_VERSION=12.18.3
+
+FROM node:${NODE_VERSION}
+ARG version=latest
+WORKDIR /home/theia
+ADD $version.package.json ./package.json
+ARG GITHUB_TOKEN
+RUN yarn --pure-lockfile && \
+    NODE_OPTIONS="--max_old_space_size=4096" yarn theia build && \
+    yarn theia download:plugins && \
+    yarn --production && \
+    yarn autoclean --init && \
+    echo *.ts >> .yarnclean && \
+    echo *.ts.map >> .yarnclean && \
+    echo *.spec.* >> .yarnclean && \
+    yarn autoclean --force && \
+    yarn cache clean
+
 FROM node:${NODE_VERSION}
 
 # Install Python 3 from source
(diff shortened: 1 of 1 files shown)


marcdumais-work:
I think there is no `theia` user in this image. If we try to stay as close as possible to original, you can just omit `--chown=theia:theia ` from the `COPY`. You may need to get rid of `USER theia` too

marcdumais-work:
> I think there is no `theia` user in this image. If we try to stay as close as possible to original, you can just omit `--chown=theia:theia ` from the `COPY`. You may need to get rid of `USER theia` too (a few lines down).

marcdumais-work:
BTW, this is a case I think where using `/home/theia` to store the app can be misleading. Normally we assume that a user exists when we see it has a home folder entry, but it's not true here. Ref: recently opened issue that asks to move the app elsewhere.

marcdumais-work:
CI now passes on the modified image

marcdumais-work:
I noticed that the full and python images did not pass CI last night, both apparently because of a python-related issue. That's new AFAIK. I'll restart Ci on this PR to see if it happens here too.

marcdumais-work:
Full finished but did not test because there were no changes in this PR... so inconclusive for that one

vince-fugnitto:
> I noticed that the full and python images did not pass CI last night, both apparently because of a python-related issue. That's new AFAIK. I'll restart Ci on this PR to see if it happens here too.

I got it to work locally again (and on CI), I'm just not sure if it was an intermittent issue or not. The main difference was using `pip3` to install the `python-language-server`: [change](https://github.com/theia-ide/theia-apps/pull/403/files#diff-db44fcad56aa1590715c9f3963aa5a86R44).


```yaml
&& pip3 install python-language-server flake8 autopep8 \
```


I also tested the extension locally and the extension does work correctly for python projects.

marcdumais-work:
> The main difference was using `pip3` to install the `python-language-server`

Kind of makes sense - the various related tools will probably drop Python 2 soon

marcdumais-work (approved):
CI is passing, except for an unrelated issue with the MacOS build. 
LGTM, but untested.

PR #408, opened 2020-09-11
java: remove 'theia-java' image

vince-fugnitto:
**Description**

The commit removes the `theia-java-docker` image from the repository as it
is outdated and currently uses an old `java8` base image, and will require
extensive updates. The `theia-full` image also contains `java` (11) so testing
should remain intact.


marcdumais-work:
I agree that this image at least needs a good update, if we were to keep it. So let's remove it for now. If we'd like to have such an image in the future, we can probably make a better one from scratch.

marcdumais-work (approved):
LGTM, thanks @vince-fugnitto

PR #434, opened 2020-10-30
swift: fix postinstall build error

vince-fugnitto:
**Description**

The following pull-request fixes the `theia-swift-docker` build caused by the missing `postinstall` script for the `sourcekit-lsp` extension: https://github.com/apple/sourcekit-lsp/blob/6511ca732e5957f6ef4f48760e11e7c1643632dc/Editors/vscode/package.json#L23-L28

```json
"scripts": {
    "vscode:prepublish": "npm run compile",
    "compile": "tsc -p ./",
    "watch": "tsc -watch -p ./",
    "createDevPackage": "npm install && ./node_modules/.bin/vsce package -o ./out/sourcekit-lsp-vscode-dev.vsix"
},
``` 

**How to test**
- the CI for the `theia-swift-docker` image should now successfully pass.

### DIFF
--- a/theia-swift-docker/Dockerfile
+++ b/theia-swift-docker/Dockerfile
@@ -4,7 +4,7 @@ RUN npm install -g vsce
 RUN git clone --depth 1 https://github.com/apple/sourcekit-lsp
 WORKDIR /sourcekit-lsp/Editors/vscode
 RUN npm install
-RUN npm run postinstall
+RUN npm run vscode:prepublish
 RUN vsce package -o ./sourcekit-lsp.vsix
 
 


vince-fugnitto:
@DukeNgn would you mind reviewing?

theia-swift-docker/Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN npm install -g vsce
 RUN git clone --depth 1 https://github.com/apple/sourcekit-lsp
```
marechal-p, line 4: Should the commit be pinned somehow? I assume this is how it broke in the first place?

marechal-p (approved):

theia-swift-docker/Dockerfile
vince-fugnitto: I think we could check out a specific tag and it’d be much more stable.

DukeNgn (approved):
The change looks good to me 👍

theia-swift-docker/Dockerfile
marcdumais-work: We're basically pulling/building their master branch, so it's possible we'll encounter some temporary issues. However any issue we see should not last very long. I am comfortable with that, so long as it's now-maintenance for us.

marcdumais-work (approved):
LGTM, thanks Vince

PR #448, opened 2020-12-04
[deb] fix image

marcdumais-work:
Minimal fix for this image. The example app would need to be updated
to support vscode extensions to be useful.

I was thinking that this may be the only example we have of packaging a 
browser Theia app, along with the `rpm` example. So we may consider 
updating the app, building a `.deb` and using it to install the app in DEBIAN
-based containers. (not in this PR though)

### DIFF
--- a/theia-deb-build-docker/package.json
+++ b/theia-deb-build-docker/package.json
@@ -4,7 +4,7 @@
     "license": "EPL-2.0",
     "description": "A custom Theia debian package",
     "engines": {
-        "node": "10.x"
+        "node": "12.x"
     },
     "dependencies": {
         "@theia/callhierarchy": "latest",
@@ -29,7 +29,7 @@
     },
     "node_deb": {
         "init": "none",
-        "dependencies": "nodejs (>= 10.0.0)",
+        "dependencies": "nodejs (>= 12.0.0)",
         "install_strategy": "copy",
         "executable_name": "theia",
         "entrypoints": {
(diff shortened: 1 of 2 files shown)


theia-deb-build-docker/package.json
```diff
@@ -4,7 +4,7 @@
     "license": "EPL-2.0",
     "description": "A custom Theia debian package",
     "engines": {
-        "node": "10.x"
+        "node": "12.x"
```
vince-fugnitto, line 7: I'm not sure if we should try to enforce a higher version of `node` like other images and the upstream repository:

https://github.com/theia-ide/theia-apps/blob/0610ffa4213c499127b372d92dc658ec726d9ca1/theia-full-docker/Dockerfile#L5

- https://github.com/eclipse-theia/theia/blob/d145b055edc304bf3c9a18cc6b57be66018fcf48/package.json#L7

It may not be an issue as other images use `12` as well:

https://github.com/theia-ide/theia-apps/blob/0610ffa4213c499127b372d92dc658ec726d9ca1/theia-swift-docker/Dockerfile#L1

vince-fugnitto (approved):
The minimal changes look good to me :+1: 
I have a minor comment regarding the `node` version.

vince-fugnitto:
@marcdumais-work fine to merge?

marcdumais-work:
yes, thanks for the reminder.

PR #457, opened 2021-01-20
Update Dockerfile

oscarnevarezleal:
Update default NODE_VERSION argument to the minimal version specified on the docs (Node.js >= 12.14.1 AND < 13)
Docs: https://github.com/eclipse-theia/theia/blob/master/doc/Developing.md#linux

### DIFF
--- a/theia-php-docker/Dockerfile
+++ b/theia-php-docker/Dockerfile
@@ -1,4 +1,4 @@
-ARG NODE_VERSION=12.18.3
+ARG NODE_VERSION=12.14.1
 FROM node:${NODE_VERSION}-alpine
 ARG version=latest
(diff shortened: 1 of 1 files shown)


theia-php-docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG NODE_VERSION=12.18.3
+ARG NODE_VERSION=12.14.1
```
vince-fugnitto, line 1: @oscarnevarezleal why is the downgrade necessary?

oscarnevarezleal: My bad, I was switching between multiple versions. No downgrade is necessary.

PR #459, opened 2021-01-29
CI: Set timeout-minutes per job to 120 minutes

DukeNgn:
## What it does
+ Sometimes, the [CI fails and the job stales](https://github.com/theia-ide/theia-apps/actions/runs/516616673) for a long time (6 hours by default). 
+ This PR sets maximum timeout minutes to 120 instead of the default 360. 

### Additional info
+ [GitHub uses 360 as default for `timeout-minutes`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes)

Signed-off-by: Duc Nguyen <duc.a.nguyen@ericsson.com>


vince-fugnitto (approved):
Looks good! 360 is too much :sweat_smile:

PR #466, opened 2021-02-18
CI: Define tag when deploy docker images

### LINKED ISSUE #465: CI fails to publish `next` images
CI currently fails to push all `next` images to DockerHub ([link](https://github.com/theia-ide/theia-apps/runs/1907106927?check_suite_focus=true#step:9:10)) because it is using the `latest` tag as the default.

DukeNgn:
Fix #465 

+ Specify the image tag (`next` or `latest`) when deploying to DockerHub. This prevents `docker push` to use the default `latest` tag for `next` images
+ The syntax is now following the format `docker push <hub-user>/<repo-name>:<tag>` from [Docker Documentation](https://docs.docker.com/docker-hub/repos/#pushing-a-docker-container-image-to-docker-hub)

Signed-off-by: Duc Nguyen <duc.a.nguyen@ericsson.com>

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -129,5 +129,4 @@ jobs:
           retry_wait_seconds: 30
           max_attempts: 3
           retry_on: error
-          command: docker push "theiaide/${{ matrix.IMAGE_NAME }}"
-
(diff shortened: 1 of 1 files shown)


vince-fugnitto (approved):
Let's give it a try, unfortunately there is no way to test without merging.

Issue #411, opened 2020-09-14
Support for arm64


XenfoMC:
I've tried running theia-full on arm64 Ubuntu (raspi4) and I get a build fail error because I'm on arm64 and not amd64. Could you guys add arm64 support so that I can host on my raspi?

vince-fugnitto:
@XenfoMC thank you for the proposal, if you are interested in providing a multi-architecture example (for arm64) of `theia-full` for example, please don't hesitate to provide a pull-request: https://www.docker.com/blog/getting-started-with-docker-for-arm-on-linux/.

XenfoMC:
> @XenfoMC thank you for the proposal, if you are interested in providing a multi-architecture example (for arm64) of `theia-full` for example, please don't hesitate to provide a pull-request: https://www.docker.com/blog/getting-started-with-docker-for-arm-on-linux/.

So by that do you mean amd64 and arm64 in the same one or a separate file?

marcdumais-work:
I read quickly the linked article (thanks @vince-fugnitto ), and I may be wrong, but I do not expect that we can achieve running an image like `theia-full` on a different arch so easily. 

I have built and run the Theia example application (Electron) on my own `raspi4` and the performance is not great, specially the startup. So I do not think a solution that uses emulation (QEMU) to run x86-64 software on this little arm CPU would work too well, practically (I expect ~5-10x slower than native binaries because emulation).

This leaves the possibility to have an image that's meant specifically for this architecture, on which we install only software meant for it. Maybe starting with a smaller image than `theia-full` would make it easier?

XenfoMC:
Maybe, all I know is that my actual computers are so old that if I host Theia it'll "blow up" my computer. I'm not really familiar with the different architectures.

Issue #130, opened 2019-02-08
ARM support


Vinno97:
Hi,

I noticed there are currently no ARM images available for Theia, but I see no technical reason for it being so. Many Dockerfiles should already be ARM-compatible, and can be adjusted with just minor changes. 
For example:
https://github.com/theia-ide/theia-apps/blob/f5117fe292c705b2ad8d2cc598b7de266b210c5a/theia-full-docker/Dockerfile#L86
can be changed to something like:
```bash
RUN curl -sS https://storage.googleapis.com/golang/go$GO_VERSION.linux-$ARCH.tar.gz | tar -C /usr/local -xzf -

```

I propose following https://blog.hypriot.com/post/setup-simple-ci-pipeline-for-arm-images/ to include the ARM builds in the CI pipeline.

---
P.S. I have tried building the images by myself on my Raspberry Pi, but all that I have tried kept getting stuck at the same place. I do not expect this to be related to it being ARM, but rather some other factor. 

```
[4/4] Building fresh packages...
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
error /home/theia/node_modules/vscode-ripgrep: Command failed.
Exit code: 1
Command: node ./lib/postinstall.js
Arguments: 
Directory: /home/theia/node_modules/vscode-ripgrep
Output:
Downloading ripgrep failed: Error: Request failed with code 401



The command '/bin/sh -c yarn --pure-lockfile &&     NODE_OPTIONS="--max_old_space_size=4096" yarn theia build &&     yarn --production &&     yarn autoclean --init &&     echo *.ts >> .yarnclean &&     echo *.ts.map >> .yarnclean &&     echo *.spec.* >> .yarnclean &&     yarn autoclean --force &&     rm -rf ./node_modules/electron &&     yarn cache clean' returned a non-zero code: 1

```

marcdumais-work:
Hi @Vinno97 

Yes, an ARM version of one or more of our docker images would be nice to have. If it could be parametrized so that we can use the same sources to build it/them, even better (e.g. like we do for "latest" vs "next" versions of the images).

> Downloading ripgrep failed: Error: Request failed with code 401

I think this error happens when the `vscode-ripgrep` npm package, during post-install, attempts to download its platform-specific executable from their releases on GitHub. It's very possible that they do not have any such executable available for ARM (Linux, MacOS and Windows are covered). 

Yep, looks-like only the major "desktop" plarforms are supported: 
[...]

PR #471, opened 2021-04-03
Changes to theia-full-docker to run on arm64

dm4tze:
Hello theia-ide team,

I've made some changes to your theia-full-docker Dockerfile to get it running on an arm64 machine (Raspberry PI4 #411 #130).
Since some of these changes will affect builds for other architectures I want to discuss my solution with you. 
If we come to a conclusion I will adjust the Dockerfile to build on the other supported architectures too.
Following changes have been made:

- Get Golang in the right architecture (I see no problems here)
- Downgraded LLVM from Version 13 to version 11 and use the distribution packets. (http://apt.llvm.org/bionic/ is not providing arm packets)
  - Do we need LLVM 13 or is version 11 ok too?
- Get CMAKE in the right architecture (I see no problems here) 
- Do not install python 2 (There seems no support anymore)
- Install .NET via tarball since there are no arm64 packets (needs some work to support other architectures)
- Install DART via zip since there are no arm64 packets.

Which of these changes are problematic? Is there a way to test all these changes to see if everything is working as expected?


theia-full-docker/Dockerfile
```diff
@@ -176,18 +183,47 @@ RUN apt-get update && apt-get -y install openjdk-11-jdk maven gradle
...
-    apt-get update && \
-    apt-get install -y \
-                       clang-tools-$LLVM \
-                       clangd-$LLVM \
-                       clang-tidy-$LLVM \
-                       gcc-multilib \
-                       g++-multilib \
-                       gdb && \
+# depending on architecture install form public LLVM PPA (Verison 13) or distirbution (Version 11)
+ARG LLVM=11
```
vince-fugnitto, line 187: @dm4tze I believe the downgrade to `11` is problematic, we want to use the latest version.

cc @marcdumais-work

dm4tze: I'm not an LLVM or C/C++ Compiler expert, so please be kind if I miss something.
For me, it looks like 11 is the latest stable version and 12 will be released next month (see https://llvm.org/).
Version 13 seems to be the snapshot version.

At the moment I will install version 13 for amd64 build and version 11 for arm64 build. (I will make that more clear in the Dokerfile)

If you insist on installing LLVM Version 13, the only solution I see at the moment is installing it from source. Would this be a feasible solution for you?

Thank you for your support!

vince-fugnitto: @dm4tze I think that's fine, if `13` is not available for your architecture (arm64) then we can use a version which is supported in that use-case.

theia-full-docker/Dockerfile
```diff
@@ -176,18 +183,47 @@ RUN apt-get update && apt-get -y install openjdk-11-jdk maven gradle
...
-    echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main" > /etc/apt/sources.list.d/llvm.list && \
-    apt-get update && \
-    apt-get install -y \
-                       clang-tools-$LLVM \
-                       clangd-$LLVM \
-                       clang-tidy-$LLVM \
-                       gcc-multilib \
-                       g++-multilib \
-                       gdb && \
+# depending on architecture install form public LLVM PPA (Verison 13) or distirbution (Version 11)
```
vince-fugnitto, line 186:
```suggestion
# depending on architecture, install from the public LLVM PPA (Version 13) or distribution (Version 11)
```

theia-full-docker/Dockerfile
vince-fugnitto, line 187: We would probably still not want to touch this value, and only set it in the arm64 architecture only.

marcdumais-work:
@dm4tze have you considered starting with one of the more modest images, that would be easier to handle? Like `theia-docker` or `theia-cpp-docker`?  The `theia-full` image is quite big (> 4GB) and seems at first look ill-suited for a rspb board, even the beefier v4 model.

marcdumais-work:
What I would love is to have an Electron-based Theia app package that works on a `rspb4`, like `VSCodium` has (they have a [bunch](https://github.com/VSCodium/vscodium/releases)).

theia-full-docker/Dockerfile
marcdumais-work: I want to add that the reason we want to bundle the latest development `clangd` and related tool is that we indirectly participate in its development and want to showcase the latest and greatest version. v11 will still work fine if nothing else is available, with some reduced features.

dm4tze: Ok, I have changed the default to 13 and only downgrade it if you build the image on arm64.

dm4tze: If you have connections to the LLVM team maybe you could ask them if they would publish the arm64 packages on their repository.

Until then, I have set the default value to 13 and only downgrade on arm build.

dm4tze:
> @dm4tze have you considered starting with one of the more modest images, that would be easier to handle? Like `theia-docker` or `theia-cpp-docker`? The `theia-full` image is quite big (> 4GB) and seems at first look ill-suited for a rspb board, even the beefier v4 model.

The idea was, first porting the full image then it should be easy to adjust the other images. At my rpi4 as backend it works without any performance issue. But I have to say, that I use a SSD as storage.

> What I would love is to have an Electron-based Theia app package that works on a `rspb4`, like `VSCodium` has (they have a [bunch](https://github.com/VSCodium/vscodium/releases)).

I don't have a rpi with installed GUI at the moment. That's why I'm using theia :)

marcdumais-work:
> I don't have a rpi with installed GUI at the moment. That's why I'm using theia :)

There is [something we're working-on](https://github.com/eclipse-theia/theia-blueprint/issues/74) that might be of interest to you. There is not that much info in the issue, but in short we're going to make available a browser app backend, in an easily packaged form. I hope the way we do it can work on raspberry pi and such, too. TL;DR the browser Theia app without Docker.

marcdumais-work:
@dm4tze  Sorry - this PR fell between the cracks for a while. I'm interested in trying it on my own raspberry pi 4. I beleive I'll need to 

Can you give a bit of details about how one goes about setting-up the necessary environment (high-level steps)? I'm ok installing a new system from scratch on a new micro-sd card, if need be. I suggest adding the info in the PR description.

theia-full-docker/Dockerfile
marcdumais-work: Awesome. If we go with one unified multi-arch image, I think we should proceed as per above when necessary. This way we maintain bleeding-edge features for the main arch, but permit other ones to evolve at their own pace and still work in the meantime, maybe with slightly reduced feature-set.

theia-full-docker/Dockerfile
```diff
@@ -196,36 +231,46 @@ RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
...
 ENV DOTNET_CLI_TELEMETRY_OPTOUT 1
 # Suppresses .NET welcome message
 ENV DOTNET_SKIP_FIRST_TIME_EXPERIENCE 1
 
-RUN curl -SLO "https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb" \
-    && dpkg -i packages-microsoft-prod.deb \
-    && rm packages-microsoft-prod.deb \
-    && apt-get update \
-    && apt-get install -y dotnet-sdk-$DOTNET_VERSION
+RUN DOTNET_ARCH= && dpkgArch="$(dpkg --print-architecture)" \
```
marcdumais-work, line 262: I think we might end-up missing some required packages, that would be marked as a dependency of the Debian package. 

Maybe better to keep installing with a Debian package for the main arch and have a different handling for arm64?

marcdumais-work:
> Which of these changes are problematic?

DOTNET and DART are the two that look suspicious to me ATM. We might need to supplement the pre-built binaries with OS--level dependencies that are expected to be present. Until we figure these out, I would be ok to keep the main architecture as-is for these languages, and having a custom setup for arm64.

If I understand correctly, GitHub workflows support arm64, to it may be possible to add that flavor of the full image to CI, by-passing the need to build on one's own board.

marcdumais-work:
@dm4tze I wanted to mention that it's highly recommended to somehow secure access to your container and its Theia app. 

You can use our "https" example image as inspiration: https://github.com/theia-ide/theia-apps/tree/master/theia-https-docker .

Without something like this, the Theia app's file-system API (among  other things) would be accessible from your LAN and become a potential "access point" e.g. for exploits that can breach your LAN, such as those that use DNS rebinding.

PR #404, opened 2020-09-09
ruby: remove ruby docker image

vince-fugnitto:
**Description**

The following pull-request removes the `theia-ruby-docker` image for the following reasons:
- the image is outdated and unmaintained
  - the image currently uses `node 8` and does not build.
  - the image is not part of CI for some time.
- the `theia-full` image contains the necessary ruby environment and [vscode-ruby](https://open-vsx.org/extension/rebornix/ruby) extension.

Signed-off-by: vince-fugnitto <vincent.fugnitto@ericsson.com>

Issue #402, opened 2020-09-08
repo: cleanup and remove unnecessary images


vince-fugnitto:
**Description**

The framework has moved towards using vscode extensions to provide language and debugging features, and consequently it should be possible to remove unnecessary images which only differ in the vscode extensions they include.

**Motivation**

1. less maintenance overall  (fewer images to maintain and manage).
2. `theia-full` already covers most languages (no need for individual images for a specific language).
3. the difference between images is minimal (main difference is vscode extensions included).
4. the images are used for reference and are not official, those who wish to create their own images can be inspired by them and they are present in the history of the repository.

**Potentially Remove:**

- `theia-rust`
- `theia-dart`
- `theia-swift`
- `theia-ruby`
- `theia-python`
- `theia-go`

cc @theia-ide/core

marcdumais-work:
> the difference between images is minimal (main difference is vscode extensions included).

I would rather say that the difference between the various Theia applications, made available in our variuos images, is minimal. So we can consider keeping some images if they bring something "special" vs `theia-full`, even if their Theia app is not special. e.g. `theia-rust` is different enough that it may be worthwhile to keep it.

vince-fugnitto:
I wanted to try and discuss what we should try to keep and what can be removed, are there images that we are particularly interested in maintaining besides `theia-docker` and `theia-full-docker`.

vince-fugnitto:
The `theia-java` should be removed, it requires a lot of changes and currently uses an old deprecated base `java8` image. In addition, the `theia-full` image already makes use of `java` (11) so there is no need for the individual image.

PR #472, opened 2021-04-04
Theia Ruby Example app

adrubesh:
This PR adds a Ruby example for Theia.


vince-fugnitto (commented):
Thank you for your contribution!

We previously had a [`theia-ruby`](https://github.com/theia-ide/theia-apps/pull/404) image as part of the repository but it was removed for the following reasons:
- The `theia-ruby` example was not useful, and did not differ too much from other examples. We also want to move towards having less overall images (https://github.com/theia-ide/theia-apps/issues/402).
- The `theia-full` image already has `ruby` support and can be used as an example for others who wish to support `ruby`.

I do not think we plan on re-adding the image at this time.

adrubesh:
@vince-fugnitto That's for the feedback. I wasn't aware of those.

PR #487, opened 2021-07-28
wip: (theia-docker) test webpack resolution

vince-fugnitto:
**What it does**

- verifies if the `webpack-sources` resolution is required to build the images by testing it on `theia-docker`.


vince-fugnitto:
The fix should be present upstream so the resolution should not longer be necessary.

Issue #485, opened 2021-07-23
latest version of theiaide/theia-python will not debug


alancummings:
The latest version (2021-07-23) of the theiaide python image on dockerhub fails to debug.
To reproduce:

1. Run `docker run --rm -it -p3000:3000 theiaide/theia-python`
2. open a browser at `http://localhost:3000`
3. create a new python file `test.py`
4. add this content:
```
import sys
print(sys.platform)
```
5. add a launch config to run the current file
6. run the launch config 
--> a message pops up: "There was an error starting the debug session, check the logs for more details."
--> the backend log has this:
```
root ERROR Error starting the debug session TypeError: is_electron_1.default is not a function
    at Object.startDebugAdapter (/home/theia/node_modules/@theia/plugin-ext/lib/plugin/node/debug/plugin-debug-adapter-starter.js:44:48)
    at DebugExtImpl.<anonymous> (/home/theia/node_modules/@theia/plugin-ext/lib/plugin/node/debug/debug.js:725:82)
    at step (/home/theia/node_modules/@theia/plugin-ext/lib/plugin/node/debug/debug.js:33:23)
    at Object.next (/home/theia/node_modules/@theia/plugin-ext/lib/plugin/node/debug/debug.js:14:53)
    at fulfilled (/home/theia/node_modules/@theia/plugin-ext/lib/plugin/node/debug/debug.js:5:58)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

- Running in Windows Docker Desktop application
- Clicking the green run button on the upper-right of the file tab _**does**_ work
- In building my own Theia IDEs, it seems that ms-python 2020.7.96456 is the _earliest_ version that will debug, and 2020.10.332292344 is the _latest_ version that will work at all ( same result as #441 )

vince-fugnitto:
@alancummings thank you for reporting the issue, please see https://github.com/theia-ide/theia-apps/issues/483, it should be fixed in the `next` version.

alancummings:
I've tested with the current "next" tagged docker container (reportedly updated on July 27):
`docker run --rm -it -p 3000:3000 theiaide/theia-python:next`
It still will not debug, but there is no longer any error output. Just, nothing at all happens. 
I suspect the version of ms-python needs to be in the range 2020.7.96456 - 2020.10.332292344 for python debugging to work in Theia (built in version is 2020.1.58038)

vince-fugnitto:
[...]

PR #492, opened 2021-08-10
update ms-python version to 2020.8.105369

alancummings:
Fixes: #485

The current docker image theiaide/theis-python will not debug python code. In building local Theia IDEs, the ms-python versions that work are in the range 2020.7.96456 to 2020.10.332292344.
Versions after 2020.8.105369, however, break the nice (but not strictly necessary) feature of showing the workspace python version in the status bar.

Updating the package.json files that include a download for the ms-python plugin.
Also, switching to use of OpenVSX as the download source.


vince-fugnitto (commented):
@alancummings thank you for the contribution, please be sure to [sign-off](https://github.com/eclipse-theia/theia/blob/master/CONTRIBUTING.md#sign-your-work) your commit so it can be accepted :+1: (the `dco` ci check is currently failing).

alancummings:
commit amended with signature

vince-fugnitto (approved):
The changes look good to me :+1: 
I confirmed that language-features and debugging work well with the version of the `.vsix`.

PR #493, opened 2021-09-02
ci: enable additional tests

vince-fugnitto:
**What it does**

The pull-request enables additional tests for docker images which were otherwise not executed.
The `top-panel` tests were simplified and should now cover cases where the top-panel is visible but no menus are present, and should hopefully help identify issues like https://github.com/theia-ide/theia-apps/issues/494.

The following images are expected to fail:
- `theia-cpp-docker` due to https://github.com/theia-ide/theia-apps/issues/494
- `theia-full-docker` due to https://github.com/theia-ide/theia-apps/issues/494

### DIFF
--- a/tests/top-panel/top-panel.ts
+++ b/tests/top-panel/top-panel.ts
@@ -20,97 +20,20 @@ export class TopPanel {
 
     public constructor(protected readonly driver: WebdriverIO.Client<void>) { }
 
+    /**
+     * Determine if the top-panel exists.
+     * @returns `true` if the top-panel exists.
+     */
     exists(): boolean {
         return this.driver.isExisting('div#theia-top-panel');
     }
 
-    openNewTerminal() {
-        this.clickMenuTab('Terminal');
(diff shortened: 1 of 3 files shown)


check_changed.sh
```diff
@@ -41,7 +41,7 @@ if [ -z "$CHANGED_FILES" ]
 then
     # nothing changed, skip building
     # this indicates to the parent script that the build can be terminated
-    echo 137
+    echo 0
```
marcdumais-work, line 44: Is this a temporary change, so that we can rebuild all images for the purpose of testing this PR?

marcdumais-work (approved):
LGTM - works well. Potentially undo what looks like a temporary change before merging!

check_changed.sh
vince-fugnitto: Right, that commit will be dropped on merge. The change was to be able to test all images, not only the ones which were modified 👍

PR #495, opened 2021-09-10
Fix full, cpp-docker and rust images, cpp-electron app

### LINKED ISSUE #494: bug: problematic `@theia/cpp-debug` extension
**Affected Images**:

- **`theia-full-docker`**
- **`theia-cpp-docker`**

**Description**

The `@theia/cpp-debug` extension is causing runtime issues for images since it still targets **es5** while the rest of the `@theia` extensions now target **ES2017**. This causes runtime issues most notably on startup where the top-level menu is missing, and contributions fail to load.

The `@theia/cpp-debug` extension needs to be updated to support **ES2017** and published so apps can make use of newer versions of the framework:
- https://github.com/eclipse-theia/theia-cpp-extensions/pull/123

**Additional Info**

- https://github.com/eclipse-theia/theia/issues/10038
- https://github.com/eclipse-theia/theia/issues/10042

marcdumais-work:
Fix theia-full, theia-cpp and theia-rust images
    
By switching to extension `cdt-gdb-vscode` instead of
using @theia/cpp-debug.
    
In addition, transition theia-full to use the new vscode
built-in extensions pack, instead of listing indivbidual
extensions.
    
Fixes #494


marcdumais-work:
@vince-fugnitto I have rebased on master so we can benefit from your recent test suite improvements, for the CI of this PR.

vince-fugnitto (commented):
The changes look good to me :+1: 
Do we not want to pin `typescript-language-features`? Next month (for `latest`) I believe the API version will be bumped and for `next` it already has been.

marcdumais-work:
> Do we not want to pin typescript-language-features? Next month (for latest) I believe the API version will be bumped and for next it already has been.

In case others are interested, we discussed this offline and concluded that we should not have to pin this plugin.

marcdumais-work:
@vince-fugnitto I have simplified the PR, only addressing the failing images and only switching to the build-ins extension pack for the full image. Can you have another look?

marcdumais-work:
I think the failure for theia@latest is unrelated to this PR.

vince-fugnitto (approved):
LGTM :+1:

PR #512, opened 2022-03-28
deb-build-docker - ubuntu needs additional packages

wuestkamp:
Otherwise the example app won't work for me when using `docker build .`


CiciDongxue:
您好，我已收到您的来信，暂时无法回复您的邮件，望见谅。

PR #518, opened 2022-10-04
Update README before archiving repo

### LINKED ISSUE #496: Ramp-down of this repository
This repository predates Theia joining the Eclipse Foundation and becoming Eclipse Theia. Initially Theia did not have support for running VS Code extensions and had to rely on Theia-specific extensions for everything, including supporting any and all languages. Given that it's required to rebuild an app if a Theia extension is added or removed, and that a single such extension not building or behaving correctly would affect the whole application (i.e. prevent it from building or running), it made sense to have separate example applications for different languages, back then. 

Nowadays, Theia supports running VS Code extensions, and it's the only way to add language support. These extensions can be added at runtime and one having an issue will not affect the support for other languages. The consequence is that the difference between the example applications we have here is the list of VS Code extensions that are installed/bundled with it. So, in essence, we have one example Theia application that can be reused in various toolchain containers. 

Until recently we did not have such an example application, that's more polished than what we have here, but we do now, and want to concentrate our efforts on it, going forward: [theia-blueprint](https://github.com/eclipse-theia/theia-blueprint). It's under our main project/license and supported by several involved companies. It's possible to use it as the base for one's own Theia-based application or to consume as-is, in packaged form: [Electron Packages (Beta)](https://download.eclipse.org/theia/latest/) for Linux, Windows and Mac are available.  

We are working on a browser version of `Blueprint`, that's intended to be a "drop-in" option, that one can add to a development container, without having to build it, like we have done so far here in this repo. That should make it way easier for those who just want a browser-accessible Theia application to use in a container or remotely, without the hassle of building it oneself

Ramp-down steps:

- [ ] Modify README to mention the ramp-down and point to the [theia-blueprint repo](https://github.com/eclipse-theia/theia-blueprint) as the official place to find a good example Theia application.
- [ ] stop publishing to Docker Hub, consider removing published images (1)
- [ ] remove most images from CI and stop maintaining them, perhaps keeping/maintaining `theia-full` for a while (best effort basis)
- [ ] Eventually stop maintaining `theia-full` and archive this repository, rendering it read-only.

(1): even if we do not remove the existing images, the Docker Hub organisation we use is a free one, that we think will be subject to their new terms of service eventually. When that happens, images older than a certain age (6 months?) will be removed automatically by them. That part is not under our control and could happen at any time. So, the difference between we removing existing images or not may end-up being a 6-months delay, after which no images will remain, either way.

marcdumais-work:
closes #496

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Collection of example cloud & desktop applications built using the [Theia](https
 
 <br />
 
-__Note: This repository is soon to be considered deprecated and will eventually be archived. [Theia Blueprint](https://github.com/eclipse-theia/theia-blueprint) is a more polished example application based on the Eclipse Theia framework, that might be of interest. For more details, see issue [#496](https://github.com/theia-ide/theia-apps/issues/496).__
+__Note: This repository is considered deprecated. [Theia Blueprint](https://github.com/eclipse-theia/theia-blueprint) and [CDT.cloud Blueprint](https://github.com/eclipse-cdt-cloud/cdt-cloud-blueprint) are more polished example applications based on the Eclipse Theia framework, that might be of interest. For more details, see issue [#496](https://github.com/theia-ide/theia-apps/issues/496).__
 
 <br />
 


CiciDongxue:
您好，我已收到您的来信，暂时无法回复您的邮件，望见谅。

vince-fugnitto (approved):
LGTM :+1:

